PR #43, opened 2015-07-29
Setup Links to open in a separate window

abishekrsrikaanth:
Setup Links to open in a separate window instead of self
Added `target="_blank"` to:
1. Github Repo Name
2. Mainfile Link
3. API Link
4. Repo Homepage Link


jimaek:
Thank you

PR #49, opened 2015-08-15
Optimize images

as-com:


jimaek:
What tool did you use for the optimization? Do you have the size savings?
Also why did you resize home-header-bg.jpg? (did you resize any others?)

as-com:
@jimaek I used a combination of pngnq, pngquant, optipng, pngcrush, zopfli, advpng, pngout, and jpegoptim (ImageAlpha and ImageOptim). I think on average I halved the size of the images.

The reason why I resized home-header-bg.jpg is because the image is blurred, and nobody would notice if it were scaled up in the browser. Plus, the file size is much smaller. No other images were resized.

MartinKolarik:
I've already used 4 lossless optimization tools. The additional savings were achieved by lossy compression, but I couldn't find any visible difference when looking at the images, so I think it's ok to merge this. The only image where I could see the difference was the resized background. We should leave the size as is.

as-com:
Ok, I returned the background image to its original size, but I still compressed it to 20 KB.

MartinKolarik:
Can you please use lower compression level on `src/public/img/features/jsdelivr-cdn-features/maximum-performance-and-uptime.png` (or exclude it)? That's the only one when I spotted a difference in quality.

as-com:
Done.

MartinKolarik:
Thank you.

PR #53, opened 2015-08-20
Added GA

jimaek:


jimaek:
Will this work with node.js? Or we need some npm module?

MartinKolarik:
Node is not a problem, but it won't work with our front-end. I'll add it.

PR #67, opened 2015-12-14
increase POP count

tomByrer:
114 objects in the [network.geojson](https://github.com/jsdelivr/jsdelivr/blob/master/network.geojson) array by my count.


jimaek:
:+1:

PR #75, opened 2016-09-27
update Cloudflare brand images

terinjokes:


MartinKolarik:
The header image dimensions are based on height, not width. Height should stay at 108px.

PR #95, opened 2017-07-24
chore(sponsors): update Algolia logo

Haroenv:
The logo was updated last December, but not yet updated here, so I did that :smile:


jimaek:
Thanks. We also have a new design coming soon, need to remember to use there the new logo.

MartinKolarik:
@jimaek and we also need to fix auto-deployment.

PR #100, opened 2017-08-30
V2

MartinKolarik:
Not done yet...


jimaek:
Looks awesome. Some feedback

- Make Top10 contain only npm
- You can remove packet as sponsor
- Can we add more dots to this map? I know its not our full map but it still looks kinda empty https://jsdelivr-com-pr-100.herokuapp.com/img/landing/network-map.png
- Where "requests served" is based on? How accurate is the number?
- I think we can remove completely the "About" page. 
- I would also make the network a separate page
- This is either broken or too empty https://dl.dropboxusercontent.com/s/e1ys25fmzuxi748/chrome_2017-08-30_09-54-06.png
- The total chart is also broken, its just a straight line

MartinKolarik:
>Where "requests served" is based on? How accurate is the number?

Average from last 30 days.

>This is either broken or too empty 

https://jsdelivr-com-pr-100.herokuapp.com/package/npm/react and add something to collection.

>The total chart is also broken, its just a straight line

The problem with that chart is that the value range is too big - from 10 to 100 000 000. There are two ways how to display the values:
 - Use a logarithmic scale (currently used) and use the same scale for all packages. This way the chart shows how popular the package is but small differences in downloads on different days are not very visible, because at this scale, 16M or 18M are almost the same value.
 - Use a different scale for each package. This way the max value for one package could be 100M and for other 10k. The advantage is that differences in values are more visible.

>On package page when you click on a folder there is no way to go back

image

But I can add `..` as well.

>Replace footer button "become a sponsor" with "our sponsors"

Wouldn't it make more sense to have "become a sponsor" there? Also, our sponsors is directly in the top menu.

jimaek:
>Use a different scale for each package. This way the max value for one package could be 100M and for other 10k. The advantage is that differences in values are more visible.

I think its best to use this. Otherwise many projects will just have straight lines which does not look good.

>But I can add .. as well.

Please do. I also think dir navigation was easier in the original design, I was confused at first. I wonder what other users will say.


>Wouldn't it make more sense to have "become a sponsor" there? Also, our sponsors is directly in the top menu.

I just wanted to promote our sponsors a bit more. If someone wants to become a sponsor they will find how to do it.

jimaek:
Also need to some SEO improvements. For example change title and description https://jsdelivr-com-pr-100.herokuapp.com/package/npm/jquery

MartinKolarik:
The current website only changes title. What do you think we should put into description? Same as title?

jimaek:
Maybe something like "Free CDN for jquery - JavaScript library for DOM operations"

>https://jsdelivr-com-pr-100.herokuapp.com/package/npm/react and add something to collection.

I think we need to show that element at all time even if its empty. It looks better than a completely empty space

MartinKolarik:
>I think we need to show that element at all time even if its empty. It looks better than a completely empty space

You are right, I actually wanted to put an info message there originally.

jimaek:
Ok, then at least lets do title.

MartinKolarik:
>I would also make the network a separate page

Not enough space in the navbar and I'd like to avoid dropdowns.

jimaek:
If we remove About then should be enough space

MartinKolarik:
No we won't and we need About more than Network in there.

jimaek:
What its going to even contain? Everything is already in Features. Network is an important feature. If it doesnt fit then at least make "features" a dropdown and Network in there

MartinKolarik:
[](https://www.jsdelivr.com/package/npm/jquery)

PR #111, opened 2017-08-31
Update total stats chart

### LINKED ISSUE #110: Stats total
chrome_2017-08-31_16-24-45

1. How about "lodash CDN stats for last 30 days" instead?
2. Remove a bit of white space?
3. The line is too empty I think. Maybe we can show dates by default without hover?

MartinKolarik:
Closes #110

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,6 +14,7 @@ global.OPBEAT_CLIENT = require('opbeat').start({
 require('./lib/startup');
 require('./lib/trace-cpu')(trace);
 
+const _ = require('lodash');
 const fs = require('fs-extra');
 const config = require('config');
 const signalExit = require('signal-exit');
@@ -28,7 +29,6 @@ const koaLogger = require('koa-logger');
 const koaETag = require('koa-etag');
 const Router = require('koa-router');
 const Handlebars = require('handlebars');
-const dedent = require('dedent-js');
 const pathToPackages = require.resolve('all-the-package-names');
 const assetsVersion = require('./lib/assets').version;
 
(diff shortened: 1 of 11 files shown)


jimaek:
Actually, how about this.
"lodash had 30.000.000 CDN hits in the last 30 days" (tweet button)

I dont think we have anywhere the combined number.

MartinKolarik:
Badge shows the total but 👍 for tweet

jimaek:
it would be even cooler if we could generate a small image on the fly with the stats that would attach to the tweet :p but thats too much

MartinKolarik:
@jimaek check now

jimaek:
That wont work for less popular projects. If someone gets 20k hits he could tweet it because its a lot to him. But if it says its #2458 project then he wont.

Lets also work on the chart near the footer in this PR.

MartinKolarik:
Well they can always modify the tweet before posting it. Showing the rank is more important and interesting for people than the number of hits IMHO.

I already updated the chart and I'd like to keep it as it is now. Styling charts is always a big pain and I think it looks exactly as it should right now.

jimaek:
>Well they can always modify the tweet before posting it. Showing the rank is more important and interesting for people than the number of hits IMHO.

Only for those with a high rank. Someone with rank 2000 wont feel the same way. But 20000 hits sounds big enough to be proud of and tweet about it.

The chart still looks like part of the background. Only the numbers on the right hint that its a chart. And easy to miss them.

I want to add more features to make it look like a real chart so people wont ignore/skip it

MartinKolarik:
As I said, they can always modify the tweet. I'm quite sure in most cases this will be used with popular projects where the rank is interesting. Also, we need to show it somewhere on the website.

>The chart still looks like part of the background. Only the numbers on the right hint that its a chart. And easy to miss them.

That's good! That was the original intention of this design, it isn't supposed to look like a regular chart.

jimaek:
>Also, we need to show it somewhere on the website.

Change it to this to see how it looks like:
lodash is the 8th most popular package on jsDelivr,

>That's good! That was the original intention of this design, it isn't supposed to look like a regular chart.

But not too much, otherwise nobody will even see it. I think we should show the dates by default https://dl.dropboxusercontent.com/s/uxtsulqzvw6xoqm/chrome_2017-08-31_19-07-14.png maybe with a doted line. And then on hover you see the hits. 
Or even just show everything by default and on hover just highlight where you are.

MartinKolarik:
>Change it to this to see how it looks like:
lodash is the 8th most popular package on jsDelivr,

Keep or revert?

jimaek:
Either keep or completely remove the rating :)

MartinKolarik:
So keep you say 😆

MartinKolarik:
Let's merge this and I'll play with the chart in a next PR.

MartinKolarik:
Actually, here's an idea: Show only if rank < 100 ?

jimaek:
Yeah that would be great

PR #116, opened 2017-09-01
Update bottom package chart

MartinKolarik:


jimaek:
@MartinKolarik Why we have 2 travis builds running at the same time?

jimaek:
The charts look correct to me

MartinKolarik:
One for commit and one for PR.

PR #134, opened 2017-09-22
feat(search): no longer manually concatenate name

Haroenv:
Hey! 

I just enabled this with a query rule on the index instead, now automatically, when there's a match with any of the concatenated names, it will be boosted as an optionalFilter. 

This means that the configuration doesn't need to be done on your side anymore. You should see the exact same results after merging this PR

screen shot 2017-09-22 at 10 23 00


MartinKolarik:
👍 thanks.

Haroenv:
Perfect, let me know if anything unexpected happens

Issue #124, opened 2017-09-07
Easier collection usage


MartinKolarik:
>extend searching to the individual package pages so that it's even easier to build up an aggregated lib link

HesyRa:
@MartinKolarik Hello. Can you please clarify for me what is the 'package pages' and how it must be done.

MartinKolarik:
Package page: https://www.jsdelivr.com/package/npm/webshim

After you select files from one package, you need to go back to the homepage if you want to search for another. The idea is that there could be a search input somewhere on the package page, and after typing there, it would bring you to the homepage with new search results.

I didn't think about where the search on package page would be yet, but maybe above "Get a badge for your package" would be fine.

HesyRa:
Thank you.

HesyRa:
![image](https://user-images.githubusercontent.com/14232439/34719599-123ee646-f544-11e7-968d-36adf58077f8.png)
![image](https://user-images.githubusercontent.com/14232439/34719610-1a59e52e-f544-11e7-81d3-0cd6138256cf.png)
![image](https://user-images.githubusercontent.com/14232439/34719617-1f9f2724-f544-11e7-8e50-d3bc1511d469.png)

MartinKolarik:
I'll leave the feedback in #150

PR #150, opened 2018-01-09
Added search block into package page

HesyRa:
https://github.com/jsdelivr/www.jsdelivr.com/issues/124

### DIFF
--- a/src/public/less/app.less
+++ b/src/public/less/app.less
@@ -19,6 +19,7 @@
 @import "components/package-file-browser.less";
 @import "components/package-header.less";
 @import "components/search.less";
+@import "components/search-compact.less";
 @import "components/slider.less";
 @import "components/stats-table.less";
 @import "components/switch-box.less";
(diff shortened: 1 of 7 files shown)


MartinKolarik:
1. The search results can't just appear in the package page. The page needs to be changed back to home as soon as user starts typing (so that there's no package page content below search results, and the URL is correct)
2. It should probably use a smaller search input in the package page because it currently takes too much space (maybe above "Get a badge for your package" ?).

Check the old version of the website to see how this should work: https://jsdelivr-com-old.herokuapp.com/

HesyRa:
@MartinKolarik Updated. Please check.

MartinKolarik:
@HesyRa thanks, that's better. I made a few improvements (check commits for details) and I think it's good to go now.

@jimaek any feedback?

HesyRa:
@MartinKolarik Thank. No questions here, everything is clear.

jimaek:
Looks good to me. Merge if the code is good to go.

Thank you both

Issue #156, opened 2018-02-07
Add loading animation


MartinKolarik:
On slow internet connection the package page takes a while to load and there's no visual indication whether it's still loading or failed.

![screenshot 2018-02-07 10 28 32](https://user-images.githubusercontent.com/6192491/35908740-cfe0f84c-0bf1-11e8-887b-087a1027dc83.png)

jimaek:
I assume its loading data.jsdelivr.com. Shouldn't most of the elements be visible but empty since its basic html?

MartinKolarik:
Everything that isn't visible in the screenshot needs data from the API.

MartinKolarik:
And the parts that need data from API are visible only after we get the data, otherwise it would look even more broken.

PR #157, opened 2018-02-07
Add loading animation

HesyRa:
https://github.com/jsdelivr/www.jsdelivr.com/issues/156

### DIFF
--- a/src/public/less/pages/package.less
+++ b/src/public/less/pages/package.less
@@ -7,18 +7,5 @@
 
 	.c-slider {
 		overflow: hidden;
-		.c-slide {
-			div:empty {
-				animation: fa-spin 2s infinite linear;
-				&:before {
-					text-align: center;
-					content: "\f021";
-					font-family: FontAwesome;
-					font-size: 50px;
-					display: block;
-					color: @c-gray;
-				}
-			}
-		}
 	}
 }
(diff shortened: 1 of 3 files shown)


jimaek:
Very broken https://dl.dropboxusercontent.com/s/sxant2nlzmmkivw/chrome_2018-02-07_15-45-30.png

jimaek:
It should also continue spinning until all elements load. Right now the main spinner disappeared once the title loaded but it had to continue until we had a file list.

MartinKolarik:
@HesyRa the animation will probably need to be controlled by JS (adding and removing a CSS class) and should be visible until we get [here](https://github.com/jsdelivr/www.jsdelivr.com/blob/master/src/views/components/package.html#L211).

src/public/less/pages/package.less
```diff
@@ -7,5 +7,18 @@
 
 	.c-slider {
 		overflow: hidden;
+		.c-slide {
```
MartinKolarik: Change this to `.c-package` (and move the code to `less/components/c-package` to fix the issue @jimaek mentioned.

HesyRa: we are thinking of trying other method. thnx

jimaek:
@MartinKolarik Please check

PR #158, opened 2018-02-12
tabular numbers in home screen stats

Haroenv:
This looks really pleasing: 

Before:

After:

### DIFF
--- a/src/public/less/components/landing-bottom.less
+++ b/src/public/less/components/landing-bottom.less
@@ -75,6 +75,7 @@
 			background-color: @c-orange-bg;
 			padding: 5px 30px;
 			border-radius: 35px;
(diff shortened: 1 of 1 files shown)


MartinKolarik:
Interesting. What browser is this? I've never seen it moving like that.

Haroenv:
That's in Safari

Haroenv:
Ah sorry, it was showing because I set up my default font to be San Fransisco which has a bigger difference between the numbers. If I remove that rule it's less noticeable, but still slightly jumpy without the rule @MartinKolarik

MartinKolarik:
Ok that makes sense, I was just surprised because it looked really bad in that picture. Thanks.

PR #172, opened 2018-05-20
add button to copy sri hash to clipboard

### LINKED ISSUE #171: Make it easier to copy the SRI
jsdelivr/jsdelivr#18072

tjenkinson:
image

The UI might need tweaking but it's a start :)

closes #171


MartinKolarik:
Thanks for the PR @tjenkinson! Could you try changing it to a dropdown, similarly to [cdnjs](https://cdnjs.com/libraries/d3)? No need to make all the options work yet, I'd just like to see how it looks. Also add this for the default file please.

MartinKolarik:
There's [this](https://github.com/jsdelivr/www.jsdelivr.com/blob/master/src/views/components/version-dropdown.html) but you can just use bootstrap html and css classes directly, no need to wrap it into a Ractive component if it doesn't have any additional features.

src/public/less/components/package-file-browser.less
```diff
@@ -70,16 +70,14 @@
...
-					margin-right: 20px;
-					margin-top: 3px;
-					cursor: pointer;
+			.actions-dropdown {
+				float: right;
+				margin-right: 20px;
+				margin-top: 3px;
+				cursor: pointer;
+
+				.fa {
```
MartinKolarik: This needs to be outside `.actions-dropdown` because it also applied to file icons on the left.

MartinKolarik:
Looks good. I think we can add "Copy HTML" and "Copy HTML with SRI" too and make it work. For the last file (or last two/three files after the new dropdown entries are added) we'll need to add the `dropup` class to fix this:

image

MartinKolarik:
No problem. I don't have the time right now either, so feel free to get back to it later or I'll finish it on the next weekend probably.

MartinKolarik:
Thanks for a well done PR @tjenkinson. Did you have previous experience with Ractive? I made a few adjustments and it should be live in a few minutes.

MartinKolarik:
@jimaek CF seems to be completely ignoring query string for css/js so it doesn't update.

src/views/components/landing-top.html
```diff
@@ -14,7 +14,7 @@ <h2>free, fast, and reliable</h2>
 					{{#if headingMessage === 0}}
 					<div slide-in-out>serving ~ {{_.formatHits(networkStats.hits.total)}} billion requests / month</div>
 					{{elseif headingMessage === 1}}
-					<div slide-in-out>serving ~ {{_.formatMB(networkStats.megabytes.total)}} TB / month</div>
+					<div slide-in-out>serving ~ {{_.formatMB(networkStats.megabytes.total)}} TiB / month</div>
```
tjenkinson: is this meant to be terabytes when it's `megabytes.total`?

src/views/components/copy-dropdown.html
```diff
@@ -0,0 +1,33 @@
+<div class="c-copy-dropdown btn-group {{#if ~/dropup}}dropup{{/if}}">
+	<span data-toggle="dropdown">
+		<i class="fa fa-clipboard" aria-hidden="true"></i>
+		<i class="fa fa-angle-down" aria-hidden="true"></i>
+	</span>
+
+	<ul class="dropdown-menu dropdown-menu-right">
+		<li><a as-clipboard="undefined, undefined, '.c-copy-dropdown'" data-clipboard-text="{{url}}">Copy URL</a></li>
+		<li><a as-clipboard="undefined, undefined, '.c-copy-dropdown'" data-clipboard-text="{{hash}}">Copy SRI</a></li>
```
tjenkinson: This no longer contains the `sha256-` prefix. Maybe this is what you want, but I noticed cdnjs does include it

src/views/components/copy-dropdown.html
```diff
@@ -0,0 +1,33 @@
...
+	const clipboard = require('../../public/js/decorators/clipboard');
+	const { buildIntegrity, buildHtml, canBuildHtml } = require('../../public/js/utils/build-links');
+
+	component.exports = {
+		decorators: {
+			clipboard,
+		},
+		data () {
+			return {
+				buildIntegrity,
```
tjenkinson: I don't think this is actually used anymore (unless you bring the prefix back)

src/views/components/package-file-browser.html
```diff
@@ -65,7 +65,7 @@
 
 					<input type="checkbox" class="switch-inline" id="{{name}}" checked="{{getCollectionIndex(~/collection, ~/type, ~/name, ~/version, '/' + pathS + name) !== -1}}" twoway="false">
 					<label for="{{name}}" on-click="@this.toggleCollectionFile('/' + pathS + name)" as-tooltip="'Add to collection'"></label>
-					<i class="fa fa-clipboard" aria-hidden="true" as-clipboard data-clipboard-text="https://cdn.jsdelivr.net/{{~/type}}/{{~/name}}@{{~/version}}/{{pathS}}{{name}}"></i>
+					<c-copy-dropdown url="https://cdn.jsdelivr.net/{{~/type}}/{{~/name}}@{{~/version}}/{{pathS}}{{name}}" hash="{{hash}}" dropup="{{@index && ../length - @index <= 4}}"></c-copy-dropdown>
```
tjenkinson: minor thing, but when it's shown above the mouse, the tooltip is still visible and covering it, and then it causes a lot of flickering when hovered over. Maybe the tooltip could have `pointer-events: none;`?

src/views/components/landing-top.html
MartinKolarik: Yes, I renamed the function to `formatMiB2TiB` to make it more clear what it does.

src/views/components/copy-dropdown.html
MartinKolarik: Oversight on my part, fixed.

src/views/components/package-file-browser.html
MartinKolarik: 👍 the tooltip was also covering part of the text so I moved it to the bottom when `dropup=true`

PR #181, opened 2018-06-28
Gh drgy

Drgy:


MartinKolarik:
Let's make it look like this:

image

src/views/components/package-header.html
```diff
@@ -69,6 +70,18 @@ <h1 class="package-name">{{name}}</h1>
 		</div>
 	</div>
 	{{/with}}
+	{{else}}
```
MartinKolarik: `{{#with package}}{{else}}{{/with}}` would work I think.

src/views/components/package.html
```diff
@@ -220,7 +220,7 @@ <h3 class="small-headline total-headline">
 						this.set('version', data.versions[0]);
 					}
 				}).catch((error) => {
-					if (error.status === 404) {
+					if (error.status === 404 && this.get('package') !== object.empty()) {
```
MartinKolarik: What's `object.empty()` ?

PR #184, opened 2018-07-30
+ package.json `browser` priority

tomByrer:
Am I assuming this correct?

Having a more general `browser` field in `package.json` may be more desirable for some package maintainers than `jsdelivr`.


MartinKolarik:
I changed the wording a bit but 👍

tomByrer:
I like "(ordered by priority)" better also.
cheers

PR #235, opened 2019-03-24
Updated SVG files of the logo and rendered fresh PNGs aswell.

metaa:
Hey there guys, I cometh in peace.

I had a look at the SVG files of your logo and noticed some pretty ugly things.

image

With this PR I would be replacing the yellow crest part of your logo with a rebuilt version so that it's all nice and round.

**Comparison:**
Comparison

For completion, here's also the source shapes I used to create the new version:
[jsDelivr_shapes_source.zip](https://github.com/jsdelivr/www.jsdelivr.com/files/3000608/jsDelivr_shapes_source.zip)

The only thing that would be left for optimization is the wordmark, because the letters also have some ugly edges (the `J`, `V`, and `R` are notable examples).
But I don't know what font you guys used 😛

# To-Dos

- [ ] **After merging:** Replace the logo [on Wikipedia](https://en.wikipedia.org/wiki/File:JsDelivr_Logo.svg).


jimaek:
Thank you for this! It looks great

Issue #260, opened 2019-05-26
Design improvements


jimaek:
I plan to talk with the designer about making some improvements. Bellow are mine but feel free to add your own and I will pass them:

1. [x] Too hard to find a link to our Github on the website. Need to find a good place for it. *Added to the top menu. Lets add Twitter there as well and remove both from footer*.

2. [x] We have a really cool infographic but its hidden behind a very small "how it works" button. Need to make it more prominent and visible. *See design below*.

3. [x] ~Maybe make the "network" page a separate page~. Show a bigger map and move all text below. *See the design below*.

4. [x] Remove about, tos and privacy from the features page. *Move all to footer*.

5. [x] Include infographic in a better way in the Features pages. *See the design below*.

6. Design purge cache tool page. Probably an input field asking for a list of URLs to purge, with "Purge Now" button and a captcha + ratelimit per IP on our side. After purge show pretty output of our results json: *Will be done in a separate issue*.

```
{
"fastly": [
{
"status": "ok",
"id": "20727-1558573413-986072"
},
{
"status": "ok",
"id": "20726-1558577588-1041027"
},
{
"status": "ok",
"id": "20726-1558577588-1041028"
},
{
"status": "ok",
"id": "20726-1558577588-1041029"
},
{
"status": "ok",
"id": "20734-1558573890-1049626"
},
{
"status": "ok",
"id": "20735-1558578641-958658"
}
],
"maxcdn": {
"code": 200
},
"cloudflare": true,
"quantil": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<response><message>success</message></response>"
}
```

MartinKolarik:
> Remove about, tos and privacy from the features page

Where do you want to link them? They are already not easy enough to find.

jimaek:
In the footer as is the standard for such pages. Forgot to include that

jimaek:
![Homepage](https://user-images.githubusercontent.com/1834071/60545134-a4bbea80-9d1a-11e9-89c5-942ec7fb1b02.png)
![Network](https://user-images.githubusercontent.com/1834071/60545136-a4bbea80-9d1a-11e9-9218-068e449c52f1.png)
![Purge cache](https://user-images.githubusercontent.com/1834071/60545158-af767f80-9d1a-11e9-897f-86cda09098f1.png)
[...]

PR #279, opened 2019-08-01
#260 follow-up (closes #278)

### LINKED ISSUE #278: #260 follow-up
1. The stackpath logo looks soooo bad. They don't officially provide this but I suggest we use the inverted version (at the very least on the homepage and https://www.jsdelivr.com/foundationcdn):

stackpath-logo-reversed-screen

[stackpath-logo-reversed-screen.zip](https://github.com/jsdelivr/www.jsdelivr.com/files/3456275/stackpath-logo-reversed-screen.zip)

2. Map marker colors are too similar. We can keep the new SVGs but should use the old colors.

Drgy:


MartinKolarik:
@Drgy foundation still looks bad.

PR #297, opened 2019-12-10
Update Cloudflare Data Center Map

SukkaW:
The data of Cloudflare data center can be found at: https://www.cloudflare.com/data/current-pops.csv , which is used to render [Cloudflare Network Map](https://www.cloudflare.com/network/). I have removed `CN` PoPs from the data, turned it into formatted json, and bring up this PR.

### DIFF
--- a/src/public/js/map/map.json
+++ b/src/public/js/map/map.json
@@ -1,653 +1,888 @@
 [
 	{
-		"title": "Cloudflare - Amsterdam",
-		"latitude": 52.2452157,
-		"longitude": 4.645167900000001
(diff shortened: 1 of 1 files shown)


MartinKolarik:
Thank you for the PR. The main source file for this data is actually https://github.com/jsdelivr/www.jsdelivr.com/blob/master/data/map.txt, could you update that instead? After updating the `.txt`, run `node bin/geocode` and that script will update the JSON map.

SukkaW:
Superseded by #298

PR #298, opened 2019-12-10
Update Cloudflare Data Center Map

SukkaW:
Continuation of #297.
I have bring up a `/bin/cloudflare.js` which could fetch cloudflare data center list, parse, filter out PoPs in China, sort and output using `console.log`.

The `map.txt` has now updated to the output of `cloudflare.js` and `map.json` has been updated as well.


bin/cloudflare.js
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env node
+const fetch = require('node-fetch-retry');
+const parseCsv = require('csv-parse/lib/sync');
+
+const result = [];
+
+fetch('https://www.cloudflare.com/data/current-pops.csv')
```
MartinKolarik: Please use `got` library here instead:

```
got('https://www.cloudflare.com/data/current-pops.csv').then(res => res.body);
```

package.json
```diff
@@ -17,6 +17,7 @@
 		"config": "^3.2.4",
 		"config-mapper-env": "^1.0.1",
 		"cookie": "^0.4.0",
+		"csv-parse": "^4.8.2",
```
MartinKolarik: Move both to dev deps.

PR #306, opened 2020-02-06
Small updates

jimaek:
- Added links to new site
- Removed email
- Removed links to cdn/dnsperf

### DIFF
--- a/src/public/less/components/footer.less
+++ b/src/public/less/components/footer.less
@@ -9,7 +9,7 @@
 	}
 
 	.footer-row-2 {
-		margin-top: 60px;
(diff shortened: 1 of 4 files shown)


MartinKolarik:
We should keep the email.

jimaek:
Its listed in the about page and privacy policy pages and our github.

MartinKolarik:
How about something like this instead?

image

Then we can keep the name and I think it's also a better position since it's related to the "Designed, developed, and maintained by" part above.

jimaek:
Ok, lets do that. Maybe then replace "by" with "and"

Issue #311, opened 2020-04-02
Remove sponsor - Update infographic


jimaek:
- [x]  Remove perfops logos everywhere
- [x]  Remove js tag
- [x]  Remove js tag call to action, but leave the benchmark
- [x]  Update infographic (need to discuss how)

jimaek:
Regarding the infographic:

- We need to remove perfops completely.
- We then change the lines to show that the DNS providers route the traffic themselves without a central load-balancer
- Then we need to design and 2 new elements in the center between DNS. 1. NS1 shown as RUM data provider(aka source) 2. Our own module that pulls this data and analyzes it and then makes decisions
- Then we show lines connecting our data module to both DNS providers showing that it pushes decision data to them directly

@MartinKolarik What do you think? I can ask the designer to update it if you agree on this logic

MartinKolarik:
He can try... the source file is here: https://www.dropbox.com/s/iqiq7podr3erzfi/infographics%20v2.psd?dl=0

I'm also not sure if our origin locations are up to date, and we're planning to switch from Amazon to DO soon, so these things should be updated as well.

jimaek:
New files attached
[infographic.zip](https://github.com/jsdelivr/www.jsdelivr.com/files/4635960/infographic.zip)

MartinKolarik:
Should be exported as png, not jpeg.

PR #321, opened 2020-04-28
Remove PerfOps as a sponsor

BirkhoffLee:
#311 
Please check the file diff


jimaek:
Need to also remove this part as well https://dl.dropboxusercontent.com/s/j5x374cmclm6ije/chrome_qYu67PKanq.png

BirkhoffLee:
Done @jimaek

jimaek:
LGTM

BirkhoffLee:
Looks like the entire RUM section has to be removed.

jimaek:
Maybe don't remove but hide it? We later should have an API of some kind to show the same data. So we could re-use the same design

MartinKolarik:
@BirkhoffLee probably comment out this section in `landing-bottom` then and add a comment referencing this issue:

```
<section class="rum">
	<div class="container">
		<c-rum></c-rum>
	</div>
</section>
```

I think this is the only place where it's used.

PR #326, opened 2020-04-28
Legacy url.parse is deprecated; switching to new URL constructor

BirkhoffLee:
Respective documentations: https://nodejs.org/api/url.html#url_constructor_new_url_input_base

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -337,7 +337,7 @@ server.use('/blog', (req, res, next) => {
 /**
  * Proxy blog requests to ghost.
  */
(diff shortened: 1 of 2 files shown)


MartinKolarik:
This isn't quite the same, which is why I didn't update it yet. See https://jsdelivr-com-pr-326.herokuapp.com/blog/ vs https://www.jsdelivr.com/blog/

MartinKolarik:
Looks better. There was still an issue with one form of relative URLs but I fixed that.

PR #328, opened 2020-05-17
copy update

jimaek:
Some copy changes

### DIFF
--- a/src/views/components/landing-top.html
+++ b/src/views/components/landing-top.html
@@ -10,8 +10,8 @@
 	<div class="content">
 		<div class="container">
 			<section class="heading">
-				<h1>A free CDN for Open Source Projects</h1>
-				<h2>Fast, global, reliable and automated</h2>
+				<h1>A free CDN for Open Source</h1>
+				<h2>fast, reliable, and automated</h2>
 
 				<div class="heading-message">
 					{{#if headingMessage === 0}}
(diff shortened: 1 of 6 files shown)


src/views/components/landing-top.html
```diff
@@ -53,6 +53,7 @@ <h2>free, fast, and reliable</h2>
 							<c-slider active="{{activeSlide}}" direction="vertical" distance="500">
 								<c-slide>
 									<div class="docs-code-slide">
+										<p class="docs-comment">// NPM CDN</p>
```
MartinKolarik, line 56: These look ugly and just duplicate the text directly below.

jimaek: Its for SEO. Its the best and most relevant place I could think to add keywords

src/views/r-page.html
```diff
@@ -10,11 +10,11 @@
 		<head>
 			<meta charset="utf-8">
 			<meta name="viewport" content="width=device-width, initial-scale=1">
-			<meta name="description" content="{{ description || 'A free, fast, and reliable Open Source CDN for npm and GitHub with the largest network and best performance among all CDNs. Serving more than 40 billion requests per month.'}}">
-			<meta name="keywords" content="jsdelivr, cdn, npm, github, js, javascript, free, open source, oss">
+			<meta name="description" content="{{ description || 'A free, fast, and reliable CDN for open source projects, deno, npm and GitHub with the largest network and best performance among all CDNs. Serving more than 80 billion requests per month.'}}">
```
MartinKolarik, line 13: "for open source projects, deno, npm" etc doesn't make much sense, if you don't want to call it open source cdn, we can just drop that part and let it be "reliable CDN for deno, npm, ..."

jimaek: But we are not limited only for them. We support any open source project with our custom configs.

MartinKolarik: Maybe but that's not the main focus, makes the text too complicated, and in no way explains what it means.

jimaek: how so? i think "a free cdn for open source projects" is pretty clear

src/views/components/landing-top.html
```diff
@@ -10,8 +10,8 @@
 	<div class="content">
 		<div class="container">
 			<section class="heading">
-				<h1>jsDelivr – Open Source CDN</h1>
-				<h2>free, fast, and reliable</h2>
+				<h1>A free CDN for Open Source Projects</h1>
+				<h2>Fast, global, reliable and automated</h2>
```
MartinKolarik, line 14: I also don't like this, both are too long. Maybe something like this

image

although I still like the current one better.

jimaek: Ok lets use your shorter version

src/views/r-page.html
MartinKolarik: Yes but when you mix it with the other terms it gets long and "open source projects" doesn't really fit in there "CDN for open source projects, deno, npm, and GitHub"

jimaek: Maybe its not ideal but I dont see any other way. Our site is not very clear or Google friendly. 
This way the user will just look at "CDN for" and then the keyword relevant to him and understand that it fits his need.

MartinKolarik: We have open source all over the place, including the main headline, it doesn't need to be everywhere.

jimaek: which always refers to us, not projects we host. Also this description is for Google where other mentions of oss are not visible

src/views/components/landing-top.html
MartinKolarik: Rephrased and moved to a better place.

PR #329, opened 2020-05-20
Update purge frontend to match API

### LINKED ISSUE #274: Add purge tool
A new page at https://www.jsdelivr.com/tools/purge. Design at https://github.com/jsdelivr/www.jsdelivr.com/issues/260#issuecomment-509980806. Limit to 10 links. Ask for API access details in a PM.

BirkhoffLee:
closes #274


src/views/pages/tools/purge.html
```diff
@@ -1,23 +1,186 @@
...
+				return;
+			}
+
+			paths = paths.split(',');
+
+			// If there are more than 10 paths, use the captcha endpoint
+			let apiEndpoint = 'https://purge.jsdelivr.net';
+			let postBody = { path: paths };
+
+			if (paths.length > 10) {
```
MartinKolarik, line 146: This should be used always.

BirkhoffLee: Do you mean always use the CAPTCHA endpoint?

BirkhoffLee:
The reCAPTCHA callback doesn't work. Also might need to add a test for the purging feature itself.

src/views/pages/tools/purge.html
MartinKolarik: Yes.

MartinKolarik:
> The reCAPTCHA callback doesn't work.

That's possible, as I said, update the code as needed, it is far from perfect.

MartinKolarik:
> Also might need to add a test for the purging feature itself.

That's not going to work because of the captcha, we can keep just the one existing test.

BirkhoffLee:
Need some help regarding this PR. There seems to be an issue with reCAPTCHA callback. The API call and endpoint are fine.

jimaek:
@BirkhoffLee How is this going? With the latest API changes maybe this PR also need an update? 
Check it out and let me know if you need anything from :)

BirkhoffLee:
yeah I'll take a look asap. really busy IRL

BirkhoffLee:
we should improve the backend so that it always return true/false instead of raw results from cdn api

jimaek:
OK, but for /captcha only to make sure we don't break anything for existing users.

MartinKolarik:
The current output is a mess and not specified anywhere. IMO feel free it change it for all responses/endpoints.

jimaek:
Its not documented but there are lots of people using it. Potentially some could be parsing the output

MartinKolarik:
If someone was checking specific properties in the response we just broke that yesterday by replacing maxcdn with bunny.

jimaek:
Lol ok then

jimaek:
the purge table looked strange so I asked a designer to improve it. What do you think?

Purge cache-3


PM me to send you the source files if needed. Also the current tool breaks if Quantil returns a rate-limit error

BirkhoffLee:
UI looks good enough to me right now. However IMHO it makes more sense to separate URLs by newlines.  
Do you mean the backend code can't properly mark request as failed when Quantil returns a rate-limit error?

jimaek:
I meant the table with the result. Currently its not great I think.

>Do you mean the backend code can't properly mark request as failed when Quantil returns a rate-limit error?
Yes, the whole tool breaks and shows no output at all

BirkhoffLee:
> the purge table looked strange so I asked a designer to improve it. What do you think?
> 
> Purge cache-3
> 
> PM me to send you the source files if needed. Also the current tool breaks if Quantil returns a rate-limit error

This new design kinda looks fine to me? Also if the backend failed to purge some of the URLs at some of the CDNs, the result would not be clear enough to the user. Maybe we should design it in the way that it shows the purge status at the four CDNs per URL?

Also what does the Quantil rate limit error look like? I'll have to make a change at the backend logic.

Issue #143, opened 2017-12-01
Add security badge


MartinKolarik:
Need to wait for algolia/npm-search#72.

PR #330, opened 2020-05-27
feat: adding security badge

bmvermeer:
#143 
Creating a badge based on a call to a reactive endpoint to feed the number of vulns a package has and link to a page that has more information on it

### DIFF
--- a/src/public/js/utils/http.js
+++ b/src/public/js/utils/http.js
@@ -38,8 +38,7 @@ module.exports.findProjectIssue = (owner, repo, title) => {
 	return $.getJSON(`${GITHUB_API_HOST}/search/issues`, { q: `${title} user:${owner} repo:${repo}` });
 };
 
-module.exports.findPackageVulns = (name, version) => {
-	// return $.getJSON(`${SNYK_API_HOST}/test/test`);
+module.exports.fetchPackageVulnerabilities = (name, version) => {
 	return $.ajax({
 		beforeSend: (request) => { request.setRequestHeader('Authorization', '8U@$Kh6Qs#b@9qxYB!QhtvLeD=e+?Hq$_b#5%x*t'); },
 		dataType: 'json',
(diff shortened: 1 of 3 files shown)


MartinKolarik:
Hi @bmvermeer, thanks for the PR. I made a few changes:
 - don't load this directly in [search results](https://jsdelivr-com-master-q2rtvyjjbv.herokuapp.com/?query=react), as it results in too many requests and backend errors,
 - minor style changes.

A few more issues that I found and that should be addressed:
 - https://jsdelivr-com-master-q2rtvyjjbv.herokuapp.com/package/npm/grunt says three vulns but https://snyk.io/vuln/npm:grunt@1.1.0 says no vulns,
 - ideally, the API key would be passed as a query parameter to avoid the additional OPTIONS request that is needed when using the header.

bmvermeer:
@MartinKolarik I can make a change to the backend that prevents this too many requests error.

MartinKolarik:
Even then I'm not sure we want to fire 10 (currently 20 because of preflight) requests on every keystroke. Maybe if there was at least 1 second denounce and the endpoint also used non-zero `max-age` for client-side caching.

MartinKolarik:
> non-zero max-age for client-side caching

that might be a good idea, regardless, at least for 1 hour

bmvermeer:
@MartinKolarik I will take a look tomorrow on the server-side to check the grunt page and the max-age thing.
At this time I cannot change the token to a query param, unfortunately. If that changes I will let you know.

MartinKolarik:
> At this time I cannot change the token to a query param, unfortunately. If that changes I will let you know.

Ok, for now, it doesn't make that big of a difference, but if we were to include it on the homepage, it would need to be changed. 10 extra requests are too many.

bmvermeer:
@MartinKolarik I deployed a new backend that fixed the mismatch you pointed out with grunt.

MartinKolarik:
@bmvermeer can we also do something about the max-age thing?

MartinKolarik:
Now live at https://www.jsdelivr.com/package/npm/lodash

bmvermeer:
Great!, I see that in the current version the badge will not be shown if the package has 0 vulns.
IMO it would be good to show explicitly that a package has no known vulns.
WDYT?

MartinKolarik:
We discussed with @jimaek and agreed to show a badge with a checkmark in that case, check now.

PR #336, opened 2020-07-04
Optimize SVG files

XhmikosR:
I used the latest svgo, and also made a few manual changes like lowercasing hex colors for better gzip compression. The files should still be editable, and I tried to follow pretty print where used like checkbox.svg and logo-horizontal.svg; the difference is negligible after compression anyway.

I can post my svgo config if needed.


MartinKolarik:
Thanks for bringing this up. I rerun svgo with config we already use on other projects (with similar results).

XhmikosR:
Your config is different than the one I used. You removed the viewbox, I personally keep it. I suggest that you revert your patch.

XhmikosR:
Plus you reverted my manual changes like the lowercase hex colors...

MartinKolarik:
> Your config is different than the one I used

Sure, that was the point.

Color case doesn't matter if it's consistent in individual files, which I think it is. If svgo was able to rewrite that automatically, I'd add it to the config but it's not worth maintaining manually.

XhmikosR:
Color case does matter in compression. That was the whole idea behind this PR: achieve the best possible compression while keeping readability.

XhmikosR:
You do not have to maintain anything. svgo wouldn't change the case if it's consistent, which is what my patch used. You basically reverted all my changes in a PR of mine.

MartinKolarik:
> Color case does matter in compression.

Can you explain how? I'm really curious about that.

PR #343, opened 2020-07-05
Enforce LF.

XhmikosR:
The previous solution didn't enforce LF. On Windows, CRLF was used which caused linting to fail.

Requires git >= 2.10.


MartinKolarik:
Thanks. Whether it's needed depends on your local gif config but I agree it's better to have it configured on the repository level, now that it works correctly.

PR #345, opened 2020-07-05
Update Bootstrap to v3.4.1

XhmikosR:
Need to check if there is any HTML passed in tooltips/popovers and if so adapt the config. This is a security release.


XhmikosR:
I couldn't find any issues from a quick look, but it definitely needs a close review.

PR #346, opened 2020-07-05
Tweak CI config

XhmikosR:
* Update to `actions/checkout@v2`
* Move environment variables to the root
* Add cache

The failure seems to be unrelated to this PR. Note that I don't cache the ChromiumDriver folder; we could do it later if needed.


XhmikosR:
The failure seems to be unrelated to this PR. Note that I don't cache the ChromiumDriver folder; we could do it later if needed.

PR #347, opened 2020-07-05
Specify `rel="noopener"` for external links.

XhmikosR:
See here for an explanation: https://web.dev/external-anchors-use-rel-noopener/

Also, move `rel` and `target` attributes at the end.

Refs #331 

EDIT: needs a close review in case I missed something, although I did use a couple of regex's.

### DIFF
--- a/src/views/components/collection-links.html
+++ b/src/views/components/collection-links.html
@@ -3,7 +3,7 @@
 
 	{{#each links}}
 	<div class="collection-link{{#if ~/combined && !@index && links.length > 1}} combined{{/if}}">
-		<a target="_blank" href="{{text}}">{{html}}</a>
+		<a href="{{text}}" rel="noopener" target="_blank">{{html}}</a>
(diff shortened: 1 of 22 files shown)


src/views/components/collection-links.html
```diff
@@ -3,7 +3,7 @@
 
 	{{#each links}}
 	<div class="collection-link{{#if ~/combined && !@index && links.length > 1}} combined{{/if}}">
-		<a target="_blank" href="{{text}}">{{html}}</a>
+		<a href="{{text}}" rel="noopener" target="_blank">{{html}}</a>
```
XhmikosR, line 6: Does this point to an external URL? If not we can drop the noopener addition

MartinKolarik: No, that's our page.

MartinKolarik:
> Also, move rel and target attributes at the end.

Please revert this and always put it first.

XhmikosR:
Sorry, not going to revert it. It's illogical to me.

Issue #353, opened 2020-07-06
Remove a couple of dependencies


XhmikosR:
* fs-extra
* bluebird

AFAICT we can do everything with native Promises and util.promisify. I haven't looked at the exact code, but less dependencies is always better. :)

PR #356, opened 2020-07-06
Remove fs-extra

XhmikosR:
There are still 2 instances, one in lib/startup.js and one in test/tests.js.

Feel free to push to my branch :)

BTW keep in mind that there's a `fs.promises` API starting from Node.js 10: https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_fs_promises_api

Refs #353


MartinKolarik:
Let's not. fs-extra is ok.

PR #357, opened 2020-07-06
Use `path.join` when concatenating paths

XhmikosR:
Should probably be enforced in ESLint later.


MartinKolarik:
Thanks but there's no point in doing that for `__dirname`

PR #359, opened 2020-07-06
Use `String.includes` instead of `indexOf`

XhmikosR:


XhmikosR:
Actually, wait, this requires ES6.

PR #363, opened 2020-07-12
Updated QUANTIL and Fastly Edge Locations

nirvanaball:
updated QUANTIL and Fastly Edge Locations based upon the information given on the official website on the URL given here - https://www.quantil.com/content-delivery-network/ and https://www.fastly.com/network-map


MartinKolarik:
For Quantil, we only use their China locations.

PR #364, opened 2020-07-12
Updated Edge Locations Cloudflare

nirvanaball:
Updated Edge Locations for Cloudflare and rectified minor spelling mistake.


MartinKolarik:
Sorry, I missed this one before. The correct way to update CF is to use this script: https://github.com/jsdelivr/www.jsdelivr.com/blob/master/bin/cloudflare.js

PR #378, opened 2021-02-04
Add migration links

### LINKED ISSUE #376: Link migration pages in footer
Has this been linked anywhere yet? This is such a useful feature of other CDNs, is there any reason it's not already linked on the homepage under the GitHub documentation?

_Originally posted by @mattrossman in https://github.com/jsdelivr/www.jsdelivr.com/issues/270#issuecomment-747756472_

Drgy:
closes #376

### DIFF
--- a/src/views/components/header.html
+++ b/src/views/components/header.html
@@ -29,6 +29,12 @@ Get all versions here: https://github.com/jsdelivr/jsdelivr-media
 						<ul class="dropdown-menu">
 							<li><a href="/tools/debug">Debug tool</a></li>
 							<li><a href="/tools/purge">Purge cache</a></li>
+							<li role="separator" class="divider"></li>
+							<li class="dropdown-header">Migrate from X</li>
+							<li><a href="/github">GitHub</a></li>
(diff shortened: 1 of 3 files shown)


jimaek:
Maybe "Convert from" would be better? Because you dont migrate from Github, you just convert URL formats

src/views/components/header.html
```diff
@@ -29,6 +29,12 @@
 						<ul class="dropdown-menu">
 							<li><a href="/tools/debug">Debug tool</a></li>
 							<li><a href="/tools/purge">Purge cache</a></li>
+							<li role="separator" class="divider"></li>
```
MartinKolarik, line 32: I actually meant this:
image

But your version, just without the X may be better:
image


@jimaek wdyt?

jimaek: Looks cleaner but what about SEO? Do those page have good titles, descriptions and text?

MartinKolarik: Yes, the pages themselves should be ok in that regard.

PR #382, opened 2021-03-03
new order of logos

jimaek:


jimaek:
Looks like the test is just broken, the second one passed and the preview looks good

Issue #355, opened 2020-07-06
Lazyload images


XhmikosR:
Since now native lazyload is a thing, I think it should be leveraged https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#Images_and_iframes

But first, #351 should be tackled IMHO. Also, do note that the images not in the viewport should be lazyloaded, so, not every one of them :)

PR #384, opened 2021-03-23
Add Lazyload attribute to images

rebeccawong7:
This pull request addresses issue #355, I have added lazy load attributes to images not in the initial viewport to help increase site performance.

I had some acceptance tests fail but I think it was due to issues with pages not loading quickly enough on my PC.

This is my first pull request so let me know if it's all OK.


src/views/components/infographic-banner.html
```diff
@@ -1,5 +1,5 @@
 <div class="c-infographic-banner">
-	<img src="{{@shared.assetsHost}}/img/landing/infographic-bg.png" srcset="{{@shared.assetsHost}}/img/landing/infographic-bg@2x.png 2x">
+	<img src="{{@shared.assetsHost}}/img/landing/infographic-bg.png" srcset="{{@shared.assetsHost}}/img/landing/infographic-bg@2x.png 2x" loading="lazy">
```
jimaek, line 2: For anything above the fold I dont think there is a point to do lazy loading. Here the whole point of the page is the image, so lazyloading would make a bad user experience. Please remove it

rebeccawong7: Done, let me know if there are any other images you want me to revert :)

PR #397, opened 2021-07-07
Gh 349

### LINKED ISSUE #349: Add autoprefixer and remove vendor prefixes
Just noticed it while digging in the codebase. Not super familiar with the setup, but shouldn't be too hard :)

xbpcb:
Fix #349


package-lock.json
```diff
@@ -1,8 +1,10502 @@
 {
 	"name": "www.jsdelivr.com",
 	"version": "2.0.13",
-	"lockfileVersion": 1,
```
MartinKolarik, line 4: Use npm v6. That will also fix the CI issue.

package-lock.json
```diff
@@ -4681,8 +4681,8 @@
 			"dev": true
 		},
 		"harmon": {
-			"version": "git+https://github.com/MartinKolarik/harmon.git#91f8be1e8666a266cc0ef1a7c440b647dd696d18",
-			"from": "git+https://github.com/MartinKolarik/harmon.git",
+			"version": "git+ssh://git@github.com/MartinKolarik/harmon.git#91f8be1e8666a266cc0ef1a7c440b647dd696d18",
+			"from": "harmon@git+https://github.com/MartinKolarik/harmon.git",
```
MartinKolarik, line 4685: revert this part manually

src/public/less/bootstrap/close.less
```diff
@@ -29,6 +29,6 @@
     cursor: pointer;
     background: transparent;
     border: 0;
-    -webkit-appearance: none;
```
MartinKolarik, line 32: The files in `bootstrap` are just a copy of twitter's bootstrap project and we shouldn't modify those.

PR #400, opened 2021-07-16
Redesign

xbpcb:
Redesign jsDelivr


MartinKolarik:
Please pull in the latest commits from master to fix the PR preview.

MartinKolarik:
Note here about wording. We are mixing up "GB" and "bandwidth" in the design. In all places where we refer to "requests" and "GB", it should be "requests" and "bandwidth" and the unit should be part of the value. E.g. this switch but it's in many places so use search for "GB" and fix everywhere:

image

topcoder728:
Hello, Nice to meet you. This is Steven from feelancer.com
https://www.freelancer.com/u/stevepol
I hope to work with you on your current ongoing project together and I can start the work right now perfectly.
I will boost the health of the current project and I will do my best.
Please hire me. I look forward to work with you. Thanks. :)

src/views/components/search.html
```diff
@@ -129,7 +162,7 @@
 				});
 
 				// Update results on input.
-				this.observe('page query __ready', (newValue, oldValue, keypath) => {
+				this.observe('page query type __ready', (newValue, oldValue, keypath) => {
```
MartinKolarik, line 165: You should be able to mix JS and CSS filters, so we should have separate `type` and `style` params.

src/views/components/search.html
```diff
@@ -66,6 +79,16 @@
 				page: 0,
 				query: '',
 				packages: [],
+				type: '',
+				onFilterItemClick( filter ) {
```
MartinKolarik, line 83: This isn't really a data function, so it should be on the component itself (same level as `scrollToSearch()`) and called as `on-click="@this.onFilterItemClick('esm')"`

src/views/components/search.html
MartinKolarik: Yes, just split it into two separate parameters everywhere, including search. And the query will look like `bootstrap type:-none style:-none`

src/views/components/search.html
itstarlight0808: So you mean `onFilterItemClick `function should be defined inside `onrender ` function?`

MartinKolarik: no, in the parent object, where `onrender` is

src/views/components/search.html
itstarlight0808: I see. Thanks for your advice.

src/views/components/search.html
itstarlight0808: I see. Thanks for your advice.

itstarlight0808: I see.  Thanks for your advice.

MartinKolarik:
Purge page:
 - only allow URLs on `https://cdn.jsdelivr.net` domain (accept http too)
 - the API accepts paths, not full URLs, you need to parse the path from the input and send only that
 - the expanded view doesn't match design
image
 - not sure why the whole row is clickable and changes border color on click but expands only when clicked on the "success" text:
image

probably expand on click anywhere and don't change border

 - the footer must stick to the bottom of the screen (this applies to all pages where the content is too short)
 
image

see https://www.jsdelivr.com/tools/purge, in CSS it was previously implemented by the wrapper/push elements

image

MartinKolarik:
History page:
 - some icons are blurred, please try to find svg version for all logs, where not available, make sure to use the correct 1x and 2x sizes for png - e.g., the rawgit icon has just one size and it's also wrong,
 - update the content based on this comment: https://github.com/jsdelivr/www.jsdelivr.com/issues/408#issuecomment-916949064

MartinKolarik:
Homepage:
 - on load, if there is a search query, and on pagination change, it must scroll to the top of the search results [like here](https://www.jsdelivr.com/?query=react), not top of the page

src/views/components/search.html
```diff
@@ -188,9 +191,9 @@
 			}
 		},
 		scrollToSearch () {
-			window.scrollTo({
-				top: Math.floor(document.querySelector('.search-input').getBoundingClientRect().top / 100) * 100,
-				behavior: 'smooth',
+			document.querySelector('.search-input').scrollIntoView({
```
MartinKolarik, line 194: This now scrolls to input in the middle of the page, instead of at the top.

IvanAdamovic646:
> Purge page:
> * [ ]  the API accepts paths, not full URLs, you need to parse the path from the input and send only that

Regarding urls in purge page, I need more details. For example, is it acceptable that I write the below url into input box?
`https://cdn.jsdelivr.net/abc`
And here, what can be the exact path to send the request to api? filtered path like this?
`https://cdn.jsdelivr.net/`
I want to confirm if this is correct or not.

src/views/components/search.html
itstarlight0808: I fixed scroll issue.
And accidently used `push -f` git command and restored it again.

MartinKolarik:
@IvanAdamovic646 user types `https://cdn.jsdelivr.net/abc` you send `/abc`, user types `http://cdn.jsdelivr.net/def` you send `/def`

MartinKolarik:
Package page: https://www-jsdelivr-com-pr-400.onrender.com/package/npm/react
 - [ ] `undefined` chart labels on Statistics tab
 
image

MartinKolarik:
Footer:

- [ ] looks broken on 800px width, we should either try to fit the desktop version to that size or show the mobile version up to 992

MartinKolarik:
I updated my to-do comments above with some new tasks.

MartinKolarik:
*Repost of my previous comments because GitHub now hides them*

Purge page:

 - [ ] on mobile version (width 600), the results box looks broken:
image


done:

<details>

 - [x] only allow URLs on `https://cdn.jsdelivr.net` domain (accept http too)
 - [x] the API accepts paths, not full URLs, you need to parse the path from the input and send only that
 - [x] the expanded view doesn't match design
image
 - [x] not sure why the whole row is clickable and changes border color on click but expands only when clicked on the "success" text:
image

probably expand on click anywhere and don't change border, make sure the cursor indicates the area is clickable

 - [x] the footer must stick to the bottom of the screen (this applies to all pages where the content is too short)
 
image

see https://www.jsdelivr.com/tools/purge, in CSS it was previously implemented by the wrapper/push elements

image

</details>

MartinKolarik:
Homepage:

 - [ ] the buttons should be aligned by type (i.e. npm always in forth column) 
image
 - [ ] no "load more" here, always show the full version (this was updated in the design too) 
image



Done:

<details>

 - [x] keywords https://www-jsdelivr-com-pr-400.onrender.com/?query=bootstrap-vue - show full text, not "..." but only first 15 keywords, allow it to be multiline, make sure it doesn't overflow (see https://www.jsdelivr.com/?query=bootstrap-vue)
 - [x] package description in search results should be full-width (12 columns, not 8)

 - [x] on load, if there is a search query, and on pagination change, it must scroll to the top of the search results [like here](https://www.jsdelivr.com/?query=react), not top of the page

</details>

MartinKolarik:
Please also check this on all pages: https://github.com/jsdelivr/www.jsdelivr.com/pull/400#issuecomment-1094100977

MartinKolarik:
probably wrong color of mobile me u on tap:

itstarlight0808:
> 
Total button count is 5 , isn't it? So you mean npm button should always be in fourth column . Is it ok to have a gap between first button and npm button?

itstarlight0808:
> 

aa

MartinKolarik:
@itstarlight0808 yes, fixed position for each button type, empty space when it's missing.

MartinKolarik:
@itstarlight0808 it'll use data from 2.1, updated doc and mocks here:
[api mocks.zip](https://github.com/jsdelivr/www.jsdelivr.com/files/8734253/api.mocks.zip)

MartinKolarik:
@itstarlight0808 it'll use data from 2.1, updated doc and mocks here:
 [api mocks.zip](https://github.com/jsdelivr/www.jsdelivr.com/files/8734263/api.mocks.zip)

itstarlight0808:
I already got it , but can't notice its usage.
Full API url is like this? 
https://data.jsdelivr.com/v1/stats/network/providers?type=hits&period=month

But it doesn't work. :(
Please give me a hint.
Thanks.

MartinKolarik:
@itstarlight0808 that's exactly how the URL will be but it isn't live yet, that's why you have sample data files there. You can create all functions as needed (including the URL construction logic) but instead of making real http calls just return data from that zip (copy paste the data into the JS as needed and when the API is live we'll just remove it and make real http calls).

MartinKolarik:
So for those two tables, render data from `API-data-2.1-network-providers.json` for now.

itstarlight0808:
I see. Thanks for your guide.

itstarlight0808:
Hello. @MartinKolarik 
How are you?
I have questions regarding 2.1 requests served over time   chart.
Currently, it's using 
 hitsData: {
	rank: 3,
	typeRank: 2,
	total: 6222085912,
	versions: {
		'0.0.1': {
			total: 18,
			dates: {
				'2022-03-26': 0,
				'2022-03-27': 0,
				'2022-03-28': 0,
				'2022-03-29': 9,
                                ...
Instead of this, should I have to use API-data-2.1-network-providers.json  data to draw chart?
Thanks.

MartinKolarik:
@itstarlight0808 yes the data right now is completely wrong, replace it with the file I sent.

itstarlight0808:
@MartinKolarik  OK. I got it.

itstarlight0808:
2.1 Requests served over time
/stats/network/providers?type={ hits | bandwidth }&country=CA&period=month
2.2 Requests served by country
/stats/network/countries?type={ hits | bandwidth}&period=month

I need to confirm these request url.
Did I properly write these request urls?

MartinKolarik:
yes

MartinKolarik:
@itstarlight0808 the first endpoint is now available in a preview environment:
https://data-jsdelivr-com-preview.onrender.com/v1/stats/network/providers?type=hits

There's one small difference compared to the previous spec: the keys in the object are provider codes instead of full names, you'll need to map those on the front-end. CF => Cloudflare, FY => Fastly, GC => G-Core.

itstarlight0808:
@MartinKolarik .   Then, how to deal with Bunny ?

MartinKolarik:
No, just ignore it and don't show it anywhere.

itstarlight0808:
You mean , just remove Bunny on UI everywhere?
1.1
1.2 
2.1 
..

itstarlight0808:
It's alright. "There you can hardcode the list without bunny."
You mean I have to leave bunny in cache hit rate section and can use static data for that.
right?

MartinKolarik:
> 

No, remove bunny there as well and increase the size of the other boxes, hopefully, that looks OK.

itstarlight0808:
OK.

MartinKolarik:
2.2. now available too https://data-jsdelivr-com-preview.onrender.com/v1/stats/network/countries?type=hits

itstarlight0808:
ok

itstarlight0808:
@MartinKolarik 
image

How to handle if interest rate is above than 100% and if it's minus?
And second question, what is QT provider? Let me know its full name.

itstarlight0808:
image

Regarding flags, Do I need to use only these countries you sent me?

MartinKolarik:
> Regarding flags, Do I need to use only these countries you sent me?

Show data for all countries. I see that some flags are missing, ignore that. I'll add the icons later.

MartinKolarik:
> @MartinKolarik image
> 
> How to handle if interest rate is above than 100% and if it's minus? And second question, what is QT provider? Let me know its full name.

That "progress bar" is actually not related to the number next to it. It should show the % share of the provider in the current period (provider.total / (sum of all providers) * 100) so it'll always be 0 - 100.

The number on the right is % change from the previous period so it can be any value but that's not an issue.

MartinKolarik:
QT is Quantil, it isn't used now but you'll indeed find it in older data (so it should be shown when the API returns it). Logo is still available in the repo: https://github.com/jsdelivr/www.jsdelivr.com/blob/3e40f6d05760d9aa3ff4caab239b1291bc17aff9/src/public/img/sponsors/quantil@2x.png

itstarlight0808:
@MartinKolarik 
image

How to transform these values into xxxx GB?
i.e      2343453423/ (8 * 1024 * 1024 * 1024) = xxx GB
right?

MartinKolarik:
Bandwidth is always in bytes so / 1024 / 1024 / 1024. There are already functions in the code to do that (and probably you'll need to use even larger units than GB).

itstarlight0808:
image

In the case of `not group platform versions` , is there only country breakdown for platform version?

MartinKolarik:
Yes.

MartinKolarik:
@IvanAdamovic646 @itstarlight0808 I changed the server handling a bit - in case you were starting it directly via `node src` you'll need to use `npm run start` now instead (if you use intellij, there's also a new gulp-based configuration).

MartinKolarik:
@itstarlight0808 we already include this library for autocomplete (it's used on the [esm page](https://github.com/jsdelivr/www.jsdelivr.com/blob/41f385be9ce2a9c05b7da6a39becf56c3d1b4f4c/src/views/pages/esm.html#L231)): https://github.com/algolia/autocomplete please use it for stats too and don't add `easy-autocomplete`

itstarlight0808:
> @itstarlight0808 we already include this library for autocomplete (it's used on the [esm page](https://github.com/jsdelivr/www.jsdelivr.com/blob/41f385be9ce2a9c05b7da6a39becf56c3d1b4f4c/src/views/pages/esm.html#L231)): https://github.com/algolia/autocomplete please use it for stats too and don't add `easy-autocomplete`

ok, I got it.

itstarlight0808:
@MartinKolarik 
Hello. Did you have a good weekend?
image

Let me know it's possible values.
Thanks

MartinKolarik:
1. npm/gh (display to user as npm/GitHub), the API URLs are https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages/npm and https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages/gh
2. the response will have this format so again you use the `prev` value https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages
3. the first project is 100% , the others are relative to that (so `hits / firstProjectHits`) - when pagination is used, it should be relative to the first project on the first page - so you always need to fetch data for the first page + the page user selected
4. it is algolia's autocomplete but an older version, the docs for the version we use are available at https://github.com/algolia/autocomplete/tree/master

MartinKolarik:
> @IvanAdamovic646 @itstarlight0808 please fix all linter issues to make CI pass. Most of them will be fixed automatically when you run `npm run lint:fix` but I don't want to do it myself to avoid merge conflicts with your local in-progress work.

ping on this

itstarlight0808:
> > @IvanAdamovic646 @itstarlight0808 please fix all linter issues to make CI pass. Most of them will be fixed automatically when you run `npm run lint:fix` but I don't want to do it myself to avoid merge conflicts with your local in-progress work.
> 
> ping on this

Can you specify which files, directories should we be responsible?
As you know, I am currently working on `statistics` page and should I fix linter issues on files related to `statistics` page?

MartinKolarik:
> > > @IvanAdamovic646 @itstarlight0808 please fix all linter issues to make CI pass. Most of them will be fixed automatically when you run `npm run lint:fix` but I don't want to do it myself to avoid merge conflicts with your local in-progress work.
> > 
> > ping on this
> 
> Can you specify which files, directories should we be responsible?
> As you know, I am currently working on `statistics` page and should I fix linter issues on files related to `statistics` page?

@itstarlight0808 I think the easiest way will be if you fix it in all files and then push the changes right away - I mostly wanted to avoid conflicts with your in-progress work. Ivan is doing only small fixes and should not have any big local changes.

itstarlight0808:
> > > > @IvanAdamovic646 @itstarlight0808 please fix all linter issues to make CI pass. Most of them will be fixed automatically when you run `npm run lint:fix` but I don't want to do it myself to avoid merge conflicts with your local in-progress work.
> > > 
> > > 
> > > ping on this
> > 
> > 
> > Can you specify which files, directories should we be responsible?
> > As you know, I am currently working on `statistics` page and should I fix linter issues on files related to `statistics` page?
> 
> @itstarlight0808 I think the easiest way will be if you fix it in all files and then push the changes right away - I mostly wanted to avoid conflicts with your in-progress work. Ivan is doing only small fixes and should not have any big local changes.

Yes, I agree with you on that, but if it will take away my time to build `stats` page
I am not sure if Dmitry will understand me. He knows about this?

MartinKolarik:
> > @itstarlight0808 I think the easiest way will be if you fix it in all files and then push the changes right away - I mostly wanted to avoid conflicts with your in-progress work. Ivan is doing only small fixes and should not have any big local changes.
> 
> Yes, I agree with you on that, but if it will take away my time to build `stats` page I am not sure if Dmitry will understand me. He knows about this?

That's OK, we can't merge the PR with failing checks :) also the automated fixes will resolve 99% of the issues, only a few will need manual changes.

itstarlight0808:
> 1. npm/gh (display to user as npm/GitHub), the API URLs are https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages/npm and https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages/gh

OK, I see.  I will do that, then,  
And another question
`npm/gh (display to user as npm/GitHub), the API URLs are https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages/npm and https://data-jsdelivr-com-preview.onrender.com/v1/stats/packages/gh`
Is this all about `most popular project` api urls?
how about `trending project` api urls?

And one more question, `top platform and top browsers` api urls are also available?

MartinKolarik:
> Is this all about most popular project api URLs?

Yes, trending is not available yet but will have the exact same format.

> And one more question, top platform and top browsers api urls are also available?

Not yet but you got the sample files for those.

itstarlight0808:
> > Is this all about most popular project api URLs?
> 
> Yes, trending is not available yet but will have the exact same format.
> 
> > And one more question, top platform and top browsers api urls are also available?
> 
> Not yet but you got the sample files for those.

OK. I see. Have a good night!

MartinKolarik:
Update regarding the country flags: link directly to https://cdn.jsdelivr.net/npm/country-flag-icons@1.5.5/3x2/ it has all the flags we need. The only visual difference is they don't have rounded corners but that can be added via css.

itstarlight0808:
image

Please resolve allow-cross-origin  CORS policy...

MartinKolarik:
@itstarlight0808 are you still seeing the error or did a reload solve it?

MartinKolarik:
If it happens again, please try to open the same URL directly and post a screenshot of the response you got as well as the response headers.

itstarlight0808:
image

MartinKolarik:
You need to open it in a new tab so that these two sections are visible in the dev tools. The headers should be there normally so I need to know what response exactly you are getting:

image

image

itstarlight0808:
I resolved it with your advice. 
when I use `/v1/stats/packages/` it doesn't work.
But if I remove `/` and use `/v1/stats/packages` , it properly works.  
Whey it cause Cors error?

MartinKolarik:
Actually, I see it in the first picture. You have an extra slash at the end. ✔️ `providers?period` ❌ `providers/?period`

MartinKolarik:
The extra slash causes a redirect and redirects don't have the CORS set. I'll change that but you shouldn't really rely on the redirects anyway.

MartinKolarik:
@itstarlight0808 Since we now have `hits` and `bandwidth` data available at the same time, let's update the tooltips for **Requests served over time** and **Requests by country** to show the average file size (calculate it as `bandwidth / hits` and convert to suitable units). It can look roughly like this:
image

MartinKolarik:
@itstarlight0808 no, keep both versions, just extend the tooltips in both to show the ratio.

itstarlight0808:
> @itstarlight0808 no, keep both versions, just extend the tooltips in both to show the ratio.

Just only show ratio `Average file size    16.67 KB` with old tooltip?? And keep other things?

MartinKolarik:
Also please run `npm install` to update your setup. It now includes automated lint & fix step on commit.

itstarlight0808:
@MartinKolarik 
How are you?
Can you let me know how to contact Dmitry outside?

itstarlight0808:
@jimaek 
Hello. Sorry for painting here with this foolish issue :)
Boss, can you leave a review for me on freelancer.com
Sorry again.

PR #401, opened 2021-08-06
Update libs "clipboard" and "marked"

rebeccawong7:
Three small changes which contribute towards optimizations listed in #331 

- Async load Google Maps JS 
- Update 3rd party libs: clipboard + marked
- Remove unused markdown component


src/views/pages/network.html
```diff
@@ -21,22 +21,6 @@
 		{{/each}}
 	</style>
 
-	{{#if !@shared.isServer && !@shared.googleMapsLoaded}}
```
MartinKolarik, line 24: There's no difference here, the scripts are already dynamically inserted and therefore async. Please revert this part.

src/public/js/decorators/clipboard.js
```diff
@@ -1,5 +1,5 @@
 module.exports = (node, title = 'Copy to Clipboard', tooltipPlacement = 'top', nodeSelector) => {
-	let clipboard = new Clipboard(node);
+	let clipboard = new ClipboardJS(node);
```
MartinKolarik, line 2: This needs an update in the eslint config.

src/views/pages/network.html
```diff
@@ -109,10 +109,10 @@ <h3>Big network in China</h3>
 
 				<div class="china-banner">
 					<div class="china-bg">
-						<img width="1140" height="328" src="{{@shared.assetsHost}}/img/network/china-bg.jpg" srcset="{{@shared.assetsHost}}/img/network/china-bg@2x.jpg 2x" loading="lazy">
```
MartinKolarik, line 112: This seems to be a mistake?

src/public/js/decorators/clipboard.js
rebeccawong7: Sorry, should have caught this before submitting the PR

src/views/pages/network.html
rebeccawong7: Done! Out of curiosity where is the dynamic insertion happening from?

src/views/pages/network.html
rebeccawong7: Did one too many ctrl+z, reverted :)

src/views/pages/network.html
MartinKolarik: The whole file is a Ractive component so the rendering is handled by Ractive and this section is specifically restricted from server rendering (`!@shared.isServer`) so it happens on the client.

MartinKolarik:
Looks good now 👍

src/views/pages/network.html
rebeccawong7: Thank you! I'll look into how Mustaches is used, I haven't come across it before

PR #413, opened 2021-09-23
RUM for the website

jimaek:
I want to add perf monitoring for our website to better detect issues and track the results of our work. It will be useful when we start working on the perf tasks and the new design


MartinKolarik:
We could use RUM from Elastic too :P

jimaek:
Please no Elastic :) I want simple solutions not that nightmare of UX

PR #418, opened 2021-10-05
stale headers

jimaek:


MartinKolarik:
We have `stale-if-error=86400` on the API and I'd do the same here.

jimaek:
I updated both numbers to be the same as API

MartinKolarik:
API has different values for different content for stale and here I wouldn't do more than 10 minutes 😄

PR #421, opened 2021-11-04
Gh 420

### LINKED ISSUE #420: Fall back to npm api on direct package page access
See https://github.com/algolia/npm-search/issues/778#issuecomment-959262434. If the Algolia index returns an error when a [package page](https://www.jsdelivr.com/package/npm/validated-changeset) is accessed, request the information from npm instead.

xbpcb:
Fix #420

### DIFF
--- a/src/public/js/utils/http.js
+++ b/src/public/js/utils/http.js
@@ -47,13 +47,6 @@ module.exports.fetchPackageVulnerabilities = (name, version) => {
 };
 
 module.exports.fetchRegistryNpmPackage = (name) => {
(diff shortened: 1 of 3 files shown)


MartinKolarik:
Oh, npm doesn't send CORS headers so this doesn't actually work 🤦‍♂️

jimaek:
@MartinKolarik Should we close the PR and issue then?

MartinKolarik:
I was thinking about proxying it through our domain but algolia already shipped a partial fix so I suppose yes.

PR #427, opened 2022-01-05
Smart social media images

### LINKED ISSUE #392: Smart social media images
Let's do something similar https://github.blog/2021-06-22-framework-building-open-graph-images/

e.g. https://github.com/vercel/og-image

zarianec:
resolve #392


MartinKolarik:
I really don't like putting binaries in the repo because it's not clear where they are from, what versions, how to update them... Is there no npm package that wraps them or does the same reasonably well?

zarianec:
>Is there no npm package that wraps them or does the same reasonably well

No, converting SVG to png is not a trivial task to do especially with JavaScript.

Source code for the binaries will be in a separate repo. I can set up CI to build binaries for each platform and to attach it to releases but I'm not sure that it makes much sense.

MartinKolarik:
> Source code for the binaries will be in a separate repo.

Wait, are these custom-written? Doesn't e.g., [sharp](https://sharp.pixelplumbing.com/) do it well?

src/middleware/open-graph/index.js
```diff
@@ -0,0 +1,165 @@
...
+
+const render = async (svg) => {
+	return sharp(Buffer.from(svg))
+		.png()
+		.toBuffer();
+};
+
+module.exports = async (ctx) => {
+	try {
+		let data = await composeTemplate(ctx.params.name);
```
MartinKolarik, line 152: You're missing possible `ctx.params.scope` here.

src/middleware/open-graph/index.js
```diff
@@ -0,0 +1,165 @@
...
+		totalFormatted: formatNum(stats.total),
+		requestsChart,
+	};
+};
+
+const composeTemplate = async (name) => {
+	let [ metadata, stats ] = await Promise.all([ prepareMetadata(name), prepareStats(name) ]);
+
+	return {
+		name: metadata.name,
```
MartinKolarik, line 134: Needs handling of long names with `...`: test with e.g. `/og-image/npm/@akqa-frontline/font-config-webpack-plugin`

MartinKolarik: For scoped packages, I think we can go with two lines: first line the scope, in a smaller font (regular text or something between), the second line the name (otherwise the name, which is the more important part, would mostly get cut). If there's not enough space, we can make the description just one line for these. See if you can make this look ok and if not, ping we can ping Patryk for suggestions.

src/index.js
```diff
@@ -273,6 +274,10 @@ koaElasticUtils.addRoutes(router, [
 	}
 });
 
+koaElasticUtils.addRoutes(router, [
+	[ '/og-image/npm/:name', '/og-image/:type(npm)/:scope?/:name' ],
```
MartinKolarik, line 278:
```suggestion
	[ '/open-graph/image/npm/:name', '/open-graph/image/:type(npm)/:scope?/:name' ],
```

src/middleware/open-graph/index.js
```diff
@@ -0,0 +1,165 @@
...
+const render = async (svg) => {
+	return sharp(Buffer.from(svg))
+		.png()
+		.toBuffer();
+};
+
+module.exports = async (ctx) => {
+	try {
+		let data = await composeTemplate(ctx.params.name);
+		let svg = await ctx.render('og-pkg-template.svg', data);
```
MartinKolarik, line 153:
```suggestion
		let svg = await ctx.render('og-pkg-template.svg.ractive', data);
```

It's not a valid svg on its own and PhpStorm doesn't like it.

MartinKolarik (commented):
Looks good, just a few small issues.

src/middleware/open-graph/index.js
zarianec: PhpStorm can't render it by syntax highlighting and formatting are still available.

src/middleware/open-graph/index.js
zarianec: Something like this should work. Both scope and name could be pretty long and must be cut but in most cases this should work pretty well

image

zarianec: But we should take in account that og images will be mostly used for preview in messengers/twitter/etc. and elements can't be too small. The regular twitter preview is just about 500px width on desktop

MartinKolarik: > But we should take in account that og images will be mostly used for preview in messengers/twitter/etc. and elements can't be too small. The regular twitter preview is just about 500px width on desktop

For sure, I meant it should be either as the description text or between that size and the previous headline size. Looking at the preview here, I think I'd make the scope a little bigger.

src/middleware/open-graph/index.js
zarianec: fixed

src/middleware/open-graph/index.js
```diff
@@ -0,0 +1,165 @@
...
+	let lineOffset = 205;
+	let lineHeight = 36;
+
+	let lines = wordwrap(charsPerLine)(description.substring(0, charsPerLine * 2)).split('\n');
+
+	if (lines.length > 2) {
+		lines = lines.slice(0, 2);
+		let lastLine = lines.pop();
+
+		if (lastLine.length > charsPerLine - 3) {
```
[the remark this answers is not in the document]
zarianec: code refactored so this change is not necessary anymore

src/middleware/open-graph/index.js
zarianec: UPD with the latest changes to the image

src/middleware/open-graph/index.js
```diff
@@ -0,0 +1,180 @@
...
+
+	return {
+		requests: {
+			total: stats.total,
+			totalFormatted: formatNum(stats.total),
+			chart: processChart(records, max),
+		},
+		bandwidth: {
+			total: stats.total,
+			totalFormatted: bytes(stats.total),
```
MartinKolarik, line 138:
```suggestion
			totalFormatted: bytes(stats.total, { unitSeparator: ' ' }),
```

src/middleware/open-graph/index.js
```diff
@@ -0,0 +1,180 @@
...
+
+const composeTemplate = async (name, scope = null) => {
+	let pkg = scope ? scope + '/' + name : name;
+	let [ metadata, stats ] = await Promise.all([ prepareMetadata(pkg), prepareStats(pkg) ]);
+
+	return {
+		name: truncateString(name, 24),
+		scope: truncateString(scope, 36),
+		description: metadata.description,
+		version: metadata.version,
```
MartinKolarik, line 152: It seems the version/author texts currently don't have any truncate logic? That could be a problem too in some cases. But maybe wait with fixing this because we're considering some design changes too.

MartinKolarik: Truncate logic that is simple and works even if we go with the dynamic positioning:

Cap the first field so that it ends at 542px from the left (simply remove letters from the end till it fits).

For the second field: assume start at 582px (which is a worst-case, assuming 542 + 40px padding), the right edge where the second chart ends, calculate fixed max-width that will be applied regardless of the actual position (i.e., the first field could be short, but we won't be changing the max-width for the second field based on that, to keep this simple).

src/middleware/open-graph/fonts.js
```diff
@@ -0,0 +1,125 @@
...
+	_getFont (fontFamily) {
+		if (!this._fonts.has(fontFamily)) {
+			throw new Error('font is not loaded');
+		}
+
+		return this._fonts.get(fontFamily);
+	}
+
+	addFont (alias, fontPath) {
+		// todo: don't forget about loadSync here
```
MartinKolarik, line 21: So what to do with this? I guess sync loading is better here, maybe rename the method to `addFontSync()` and remove comment?

src/middleware/open-graph/index.js
```diff
@@ -0,0 +1,207 @@
...
+	let maxLineWidth = 760;
+	let lineOffset = 263;
+	let lineHeight = 48;
+
+	let lines = fontsProcessor.wrap(description, 'Lexend Regular', 28, maxLineWidth);
+
+	if (lines.length > 2) {
+		lines = lines.slice(0, 2);
+		let lastLine = lines.pop();
+		let lastLineWidth = () => fontsProcessor.computeWidth(lastLine, 'Lexend Regular', 28);
```
MartinKolarik, line 99: The svg size is 30 but I'm guessing that's compensated by the negative letter spacing?

zarianec: fixed

src/middleware/open-graph/fonts.js
zarianec: fixed

PR #428, opened 2022-01-19
Product page search

vdshv:
Search-compact.html is basically a copy of search.html with minor differences, so I think we could organise it better to avoid duplicated code

I'm not sure why, but after some testing I saw the existing homepage search didn't work on product page when it was addressed directly from header, but worked  perfectly when it was added in _package.html inside the header declaration

Please let me know if better approach can be used here


xbpcb (approved):
Seems ok now

jimaek:
Conflicts need solving

PR #441, opened 2022-06-30
Gh 435

xbpcb:
Removed Debug tool page.
Modified top menu content.
Added Documentation page (available by Resources -> Documentation) instead of external link.


src/public/less/pages/documentation.less
```diff
@@ -0,0 +1,337 @@
+@scrn-md: 768px;
+
+.p-documentation {
```
MartinKolarik, line 3: Not sure about this file - is it all unique styles or how much are we duplicating the policy styles? If needed, we can use a more generic class name for the shared styles.

src/public/less/pages/documentation.less
```diff
@@ -0,0 +1,337 @@
+@scrn-md: 768px;
```
MartinKolarik, line 1: I believe we already have a set of vars defined that can be imported and used.

src/public/less/pages/documentation.less
xbpcb: Terms page is broken, need to be fixed. Because the Homepage and package page is in priority i will fix Terms right after them with the usage of Docs page styles.

src/public/less/pages/documentation.less
xbpcb: You are right, fixed

PR #444, opened 2022-07-22
Fix terms page

xbpcb:
Use the same styles for both Terms and Documentation pages. Fix Terms page,


src/views/pages/terms.html
```diff
@@ -11,7 +11,11 @@
 	<c-header></c-header>
 
 	<div class="p-privacy-policy">
-		<div class="dropdown">
+		<div class="please-wait-sign {{#if @shared.auNetText && @shared.ppNetText && @shared.ppComText}} not-displayed{{/if}}">
```
MartinKolarik, line 14: I have to say I don't understand this logic at all - we're showing the loading message if all those variables **exist**? It seems to actually work but the code seems like it should do the exact opposite.

MartinKolarik (commented):
Not sure why but the main navbar is hidden until the docs load, which it shouldn't.

src/views/pages/terms.html
xbpcb: Nope, we adding the class .not-displayed which will hide the message once we got all data we need

xbpcb: > Not sure why but the main navbar is hidden until the docs load, which it shouldn't.

I will check it

jimaek: Any news on this?

xbpcb: I can't reproduce this bug no more. I've checked the code and find nothing, testing with throttling, different browsers etc. - nothing.

PR #445, opened 2022-07-25
feat: add discord links

ayuhito:
Adds a redirect for `/discord` and the Discord link where socials are shown.


ayuhito:
I do have a slight concern about it adding a little too much clutter, but that might just be me. Will make a subsequent PR for the redesign branch if this is approved.

MartinKolarik:
@jimaek wdyt? 
image

ayuhito:
Personally, I think if the esm.run part wasn't there, it would look more natural. Fortunately that's already done in the redesign branch, so it shouldn't be problematic to include it there imo.

jimaek:
Looks good to me, at least on the current site. On the new one we will need to find a better location

PR #446, opened 2022-07-26
fix: add router ignore to discord link

ayuhito:
Fixes the redirect since it seemed to be the client side router responsible for the bug.

### DIFF
--- a/src/public/js/app.js
+++ b/src/public/js/app.js
@@ -92,7 +92,6 @@ app.router.addRoute('/terms', cPP);
 app.router.addRoute('/terms/privacy-policy-jsdelivr-com', cPPCom);
 app.router.addRoute('/terms/privacy-policy-jsdelivr-net', cPPNet);
 app.router.addRoute('/terms/acceptable-use-policy-jsdelivr-net', cAUNet);
-app.router.addRoute('/discord', () => { location.href = 'https://discord.gg/by8AcrjvRB'; });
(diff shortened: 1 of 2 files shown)


MartinKolarik:
There are multiple ways to bypass the client-side router, but I have already changed it to point directly to discord. I see no reason to use the redirect here - it makes sense when you want to post the text version of the link, but here it doesn't really matter what's the URL when you just click an icon.

ayuhito:
We do want to share the link on Twitter and other places, so we do need to get the redirect working.

I was thinking to change the icon link to the redirect as there is a possible interaction of a user copying the link from the icon to share. I can alter that if you don't think that's appropriate though.

MartinKolarik:
But the redirect works, just not when you click the link on our website. Anyway, you can add `router-ignore` class to the link and then you don't have to set up a client-side route for it (and can keep the link to `/discord`).

ayuhito:
Yup, that works. I think there's a slight issue with caching though with this solution, which might affect live users. It works fine if I use an incognito tab, but the catch all still occurs with my normal tabs, even with Devtools Network cache disabled. Is that ok to push?

MartinKolarik:
Hmm ok, it indeed doesn't currently work in the live version but it's because of an unrelated caching issue. You can ignore that here, just add the class and change href.

MartinKolarik:
I think it must be on the link and not the image?

PR #449, opened 2022-08-10
Fix about history page

xbpcb:
Fix mobile and desktop styles of the About and History pages

### DIFF
--- a/src/public/less/components/header-gradient.less
+++ b/src/public/less/components/header-gradient.less
@@ -70,7 +70,7 @@
 }
 
 .p-history .c-header-gradient {
-	height: 389px;
+	height: 388px;
 }
 
 .p-unpkg .c-header-gradient,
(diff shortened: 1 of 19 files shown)


xbpcb:
Not ready yet.

xbpcb:
There is a problem with the tests, the test for the search form on the main page is currently broken due to the fact that the jsDelivr package is expected first, but currently 2 new packages have appeared and therefore the test does not pass. Will fix it later in the  separate PR

MartinKolarik:
The test issue is indeed result of some search logic changes. It's a bit unfortunate but we can just change the test.

MartinKolarik:
Adding back `-webkit-background-clip: text` because the prefix is needed in Chrome and autoprefixes doesn't add it.

MartinKolarik:
- [ ] 
image

PR #450, opened 2022-08-19
Fix network and migrating pages

xbpcb:
Fixed mobile and desktop styles for Network, Google, Github, Unpkg, Skypark, ESM.SH pages

### DIFF
--- a/src/public/less/pages/migrating.less
+++ b/src/public/less/pages/migrating.less
@@ -215,28 +215,23 @@
 				background: #f65128;
 				border: 4px solid #f6d3ca;
 				border-radius: 50%;
-				margin-bottom: 8px;
+				margin: 0 0 8px;
 				transition: background 400ms 200ms, border 400ms 200ms;
 
-				@media (min-width: @screen-sm-min) {
-					margin-bottom: 4px;
+				&.bottom {
+					margin: 8px 0 0;
 				}
 
-				&.bottom {
-					margin-top: 12px;
-					margin-bottom: 0;
+				&-disabled {
+					width: 8px;
+					height: 8px;
+					background: #cfd6dd;
+					border: none;
 
-					@media (min-width: @screen-sm-min) {
-						margin-top: 8px;
+					&.bottom {
+						margin: 12px 0 0;
 					}
 				}
-			}
-
-			.circle-disabled {
-				width: 8px;
-				height: 8px;
-				background: #cfd6dd;
-				border: none;
 			}
 
 			.line {
--- a/src/public/less/pages/migrating.less
+++ b/src/public/less/pages/migrating.less
@@ -289,7 +289,7 @@
 	.container {
 		width: 100%;
 		padding: 0 24px;
-		max-width: 1016px;
+		max-width: 1064px;
 	}
 
 	.mb-48 {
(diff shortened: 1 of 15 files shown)


src/views/pages/network.html
```diff
@@ -53,74 +53,94 @@
...
-				{{#each providers}}
-					<div class="network-control">
-						<input class="switch-inline {{this}}"
-							type="checkbox"
-							id="provider-{{this}}"
-							checked="{{enabled[@key]}}"
-							twoway="false">
-							<label for="provider-{{this}}" on-click="@this.toggleMap(@key)"></label>{{this}}
-					</div>
+				{{#each providers:idx}}
```
MartinKolarik, line 56: Don't understand the goal of the pairing here. It'll now break if we have an odd number of providers. The previous version handled that case.

xbpcb: Why it will break, it will render every pair of the array for odd and even number of elements. I paired em because the length of providers could be different and in this way it easy to style this block of controls properly with no need of manual fixing every time

src/public/less/pages/migrating.less
```diff
@@ -46,14 +46,18 @@
 		display: flex;
```
MartinKolarik, line 46: Somehow I get like 1 FPS on the animation here, any ideas? It looks terrible.

src/views/pages/network.html
MartinKolarik: image

MartinKolarik: Correct (actually I think the last line was supposed to be centered in this case but we no longer have that screen in figma):

image

src/public/less/pages/migrating.less
xbpcb: I left the animation the same as it was before my changes. I thought it was checked, i can fix it, how should it look like?

src/views/pages/network.html
xbpcb: So how it should look like?
 If the number of elements is odd, how should the last one line look? The control which takes the whole length of block and justified to the left? Or justified to the center, or in some other way?

MartinKolarik: Like the last picture but probably centered.

src/views/components/comparison-table.html
```diff
@@ -1,6 +1,13 @@
 <div class="c-comparison-table">
-	<div class="jsdelivr-wrapper"><div></div></div>
+	<div class="jsdelivr-wrapper">
```
MartinKolarik, line 2: The tooltips in the tables no longer work.

src/views/pages/network.html
xbpcb: Ok

src/views/components/comparison-table.html
xbpcb: Will check

src/views/pages/network.html
xbpcb: Like this right?

MartinKolarik: yes

src/views/components/comparison-table.html
xbpcb: Fixed

PR #453, opened 2022-08-31
Fix esm run page

xbpcb:
Fix Esm.run page styles


MartinKolarik:
1. The benchmark button must have `cursor: not-allowed` when running. 
2. Missing `';` at the end
image

PR #454, opened 2022-09-06
Fix purge page

xbpcb:
Fix Purge page styles


MartinKolarik:
- [ ] Entering more than 10 links results in a broken state, the captcha stays open, and there's no error message, but nothing happens. Probably there should be an error right away similar to this:
image

PR #455, opened 2022-09-11
feat(website): add microdata

### LINKED ISSUE #451: Add micro data annotations
Following https://github.com/jsdelivr/jsdelivr-backend/issues/323, add annotations where possible - likely the main package page but also parts of homepage/stats page where we show package-related info.

ayuhito:
Closes #451. Adds rich results for Logo and Sitelinks, and more general information for each package. 

I've been a little wary on how much is too much or if I should include data that feels irrelevant, especially after digging around on e-commerce websites like Shopify or H&M which seem to be pretty reserved.


src/views/pages/_index.html
```diff
@@ -8,6 +8,34 @@
 <link rel="ractive" href="../components/header-gradient.html" name="c-header-gradient">
 
 <r-page noYield="{{noYield}}" title="{{title}}" description="{{description}}">
+	{{#partial seo}}
+	<script type="application/ld+json">
+		[{
+			"@context":"http://schema.org",
```
ayuhito, line 14: I can include more people data here such as founders or company size. Should I do that? This particular snippet is the most impactful one out of everything else.

e.g.
image

src/views/pages/_package.html
```diff
@@ -14,6 +14,30 @@
...
+	{{#partial seo}}
+		<script type="application/ld+json">
+			{
+			"@context": "https://schema.org",
+			"@type": "WebPage",
+			"mainEntity": {
+				"@context":"http://schema.org",
+				"@type":"SoftwareSourceCode",
+				"name":"{{name}}",
+				{{#if versionP}}"version":"{{versionP}}",{{/if}}
```
MartinKolarik, line 26: You can define all this in a real JS as component data and then just do `{{{JSON.stringify(...)}}}` here

src/views/pages/_index.html
MartinKolarik: I'd say as long as there's a schema for it and the information is at least somewhat meaningful, add it.

src/views/pages/_package.html
ayuhito: Sure thing. I moved it `computed` although I'm not 100% sure if that's the right area to keep it in. Feel free to let me know if I there's a better way of doing it :+1:

src/views/pages/_package.html
```diff
@@ -84,6 +90,28 @@
...
+						'@type': 'SoftwareSourceCode',
+						'name': this.get('name'),
+						'version': this.get('versionP'),
+						'description': this.get('description'),
+						'url': `https://www.jsdelivr.com/package/npm/${this.get('name')}`,
+						'codeRepository': this.get('package.repository.url'),
+						'dateCreated': Object.values(this.get('package.versions')).at(0),
+						'dateModified': Object.values(this.get('package.versions')).at(-1),
+						'programmingLanguage': 'JavaScript',
+						'license': this.get('package.license'),
```
MartinKolarik, line 106: Looks good, except I think this one is invalid because it's supposed to be CreativeWork or URL, not a string.

ayuhito: Does package have access to the Github license URL?

MartinKolarik: No, probably just remove it.

PR #458, opened 2022-09-30
Fix custom cdn oss page

xbpcb:
Fixed styles of CustomCDNOSS and UstomCDNOSSProject pages.
Reworked chart data handling and label creation.
Label creation is not polished yet. Code requires a few updates (TODOs)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -38,6 +38,7 @@ const getRollupStream = () => rollupStream({
 	output: {
 		name: 'app',
 		format: 'umd',
+		sourcemap: true,
 		globals: {
 			algoliasearch: 'algoliasearch',
 			ractive: 'Ractive',
--- a/src/public/js/_.js
+++ b/src/public/js/_.js
@@ -404,13 +404,14 @@ module.exports = {
 				case 'month':
 					if (!resData.preparedData[dateMonth]) {
 						resData.preparedData[dateMonth] = {
-							value: valueByDateConverted,
-							name: periodMonthFormatted,
-							startsOn: date,
 							isFull: false,
+							value: valueByDateConverted,
+							startsOn: date,
+							month: periodMonthFormatted,
+							year: dateYear,
 						};
 					} else {
-						let lastDayNum = _.getAmountOfDaysByMonth(dateMonth, dateYear);
+						let lastDayNum = this.getAmountOfDaysByMonth(dateMonth, dateYear);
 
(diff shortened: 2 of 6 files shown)


MartinKolarik:
Note that for all charts:
 - [ ] the Y axis should start at zero,
 - [ ] probably use [this function](https://github.com/jsdelivr/www.jsdelivr.com/blob/146d8a62978ad591c1719bf1331d6e88b71ffb7f/src/public/js/_.js#L88) to generate the Y ticks (`linRange(maxValue, ticksCount)`)

MartinKolarik:
- [ ] I would say the year here should be shown when it changes, i.e., in January (similarly, when we should day + month, the month should be under the first day).
image

MartinKolarik:
- [ ] See how this chart works at https://www.npmjs.com/package/react
image

Can we do the same and display the date range in the headline when one is highlighted? If it's too hard, then maybe in the tooltip as "date1 - date2: hitsValue" (when grouped) and "date: hitsValue" (when not grouped). Let's test this first on one chart but then we should do it for all.

MartinKolarik:
- [ ] The labels on Y axis must be aligned to the right.

MartinKolarik:
- [ ] align 
image

MartinKolarik:
- [ ] let's try to add a link here which goes back to `/custom-cdn-oss`
image

xbpcb:
Seems ready

MartinKolarik:
> Note that for all charts:
 the Y axis should start at zero,
 probably use [this function](https://github.com/jsdelivr/www.jsdelivr.com/blob/146d8a62978ad591c1719bf1331d6e88b71ffb7f/src/public/js/_.js#L88) to generate the Y ticks (linRange(maxValue, ticksCount))

image

Both marked as done but neither seems so

MartinKolarik:
The current interaction logic (for bar highlighting and tooltips) doesn't work well with the smallest bar size because we use `intersect: true`, and the bars are very small to hover them precisely. We should use a [custom interaction mode](https://www.chartjs.org/docs/latest/configuration/interactions.html#custom-interaction-modes) which works like `nearest, axis: x` [here](https://www.chartjs.org/docs/latest/samples/tooltip/interactions.html) except with a limit of 2 pixels on each side - so it picks the closes bar if there is one within two pixels only. Otherwise, it doesn't pick anything (as with default `intersect: true` behavior). Basically, it'll add a 2px hover area around each bar.

PR #464, opened 2022-10-31
Additional fixes to package page

xbpcb:
Display downloads of files and directories of Package File Browser.
Fix Statistics tab charts behavior.


xbpcb:
Migration from ${API_HOST}/v1/package/${type}/${name} to ${STAGING_API_HOST}/v1/packages/${type}/${name} is done

src/views/components/package-file-browser.html
```diff
@@ -183,40 +203,92 @@
...
+							this.set('filesLoadingMessage', errorHtml);
+						} else {
+							this.set('filesLoadingMessage', defaultError);
+						}
+					});
+
+					// fetch package File stats to get downloads amount per file
+					http.fetchPackageFileStats(packageType, packageName, version).then((response) => {
+						let downloadsPerFile = response.reduce((downloadsPerFile, fileStats) => {
+							let fileName = fileStats.name.substr(fileStats.name.lastIndexOf('/') + 1);
```
MartinKolarik, line 244: This logic here won't work with files that have the same name but are in different directories. In the file browser you have a `path` variable which you'll need to use too to access these stats.

src/views/components/package.html
```diff
@@ -247,31 +246,6 @@
 	};
 
 	component.exports = {
-		computed: {
```
MartinKolarik, line 250: Not sure why this is removed.

src/views/components/package-file-browser.html
xbpcb: You are right. Fixed.

MartinKolarik:
1. On smaller screens the sidebar should move above
image
2. There should be a min-width and scroll bar if needed.
3. If the links are still too long, we should cut them and show `...`

See how all of the above works in the current master:
image

MartinKolarik:
Not sure why but the versions dropdown has `max-width` 320, which is correct, but it still renders much more narrow, even if it contains longer versions:

image

It should expand up to 320px width. Applies to package browser and stats tabs.

xbpcb:
#466 fixed here

PR #472, opened 2022-11-28
WIP: Fix Statistics page

xbpcb:
Everything is ready for check except the Map element (Requests served by country)


src/public/js/_.js
```diff
@@ -1018,6 +1019,8 @@ module.exports = {
 				return 'G-Core';
 			case 'QT':
 				return 'Quantil';
+			case 'BY':
```
MartinKolarik, line 1022: Import `src/public/json/net-providers.json` where you need this and don't duplicate it anywhere else.

MartinKolarik: See e.g. `src/views/components/network-provider-stats.html:96`

xbpcb: Yes, you're right, i will review my code in order to use net-providers.json, once i am done with the Map element

xbpcb: Done

xbpcb:
One more bug lefts, when user tries to switch page Limit on the Top Platform/Browsers tables or Projects Statistics tables

MartinKolarik:
Marked as done but isn't:
 - [ ]  Looks like this removed the top filter sticky behavior? See how it worked in redesign - depending on which section you're scrolled to, the period/location filters should be shown sticky at the top.

MartinKolarik:
Reposting because GH has hidden this again:
 - [x] Map tooltip colors should match Requests served over time (dark style not white)
 - [x] Show market share for browsers/platforms as reported by the API, without rounding, including decimals. UPDATE: enforce two decimals to make it more consistent, e.g. `40.90`, not `40.9`
 - [x] Numbers should be aligned to the right in most places:
   - [x] https://dl.dropboxusercontent.com/s/vx3wfgzfxw1nql6/2022-12-07_19-24-23.png
   - [x] https://dl.dropboxusercontent.com/s/pz7lueonbdodipq/2022-12-07_19-25-37.png
   - [x] https://dl.dropboxusercontent.com/s/t08c6mqz82p69l1/2022-12-07_19-28-10.png
  - [x] Trending projects table can be removed, we don't have data for it.
  - [ ] Most popular projects incorrectly shows "Records 1 - 10 of 100" instead of using backend pagination it considers only the first page with 100 results. Same issue for platforms/browsers. The pagination should use `page` and `limit` api params.
  - [x] The map scale should use the same logic for choosing the numbers as we have on other charts https://dl.dropboxusercontent.com/s/1cm9w64ixnpnhqb/2022-12-07_19-49-02.png
  - [ ] Let's add an option "Show provider breakdown" with default state on. When off, show simply one line with numbers for all providers summed up. 
image
 - [x] Tables must be re-sorted when data changes: https://dl.dropboxusercontent.com/s/g598cv1sffa7ebe/2022-12-08_18-48-29.png
 - [ ] See if we can configure the map like this: mouse scroll wheel scrolls the page by default (right now it zooms the map), to zoom, you need to use ctrl + scroll (same as Google maps works when embedded).
 - [ ] double border https://dl.dropboxusercontent.com/s/bigqd09hrl0l4yf/2022-12-09_12-47-00.png
 - [ ] bad z-indexes https://dl.dropboxusercontent.com/s/zzkuf530eqwbirj/2022-12-09_12-48-41.png, https://dl.dropboxusercontent.com/s/s4fova8162nc7yd/2022-12-09_13-11-26.png
 - [ ] Missing separator after global https://dl.dropboxusercontent.com/s/1s5apar59qrva1p/2022-12-09_13-20-09.png
 - [ ] Add continents to the country dropdown. Order: Global, separator, continents, separator, countries. Use the global flag for them. API details: https://www-jsdelivr-com-pr-400.onrender.com/docs/data.jsdelivr.com#get-/v1/stats/network
 - [ ] The flag needs styles `border: 1px solid #f5f5f5; border-radius: 3px;` here so can't be done as bg image: 
image

src/public/js/_.js
```diff
@@ -323,7 +323,7 @@ module.exports = {
 		return unit.symbol;
 	},
 
-	formatBytesWithUnit (num, unit) {
+	formatBytesWithUnit (num, unit, delimiter = ' ') {
```
MartinKolarik, line 326: There is no need for this parameter because there should *always* be a space, that's why I removed the `delimiter` option before.

xbpcb: ok

MartinKolarik:
Reposting because GH has hidden this again:
 - [ ] See if we can configure the map like this: mouse scroll wheel scrolls the page by default (right now it zooms the map), to zoom, you need to use ctrl + scroll (same as Google maps works when embedded).
 - [ ] In tables, any kind of parameter change - grouping/breakdown/per-page count... must reset page to 1.
 - [ ] Must be displayed nicely in two lines (better) or cut to fit https://dl.dropboxusercontent.com/s/nm5614rcpokw8gp/2022-12-21_00-15-15.png
 - [ ] https://dl.dropboxusercontent.com/s/a3dyt12kytqjstl/2022-12-21_16-26-36.png

PR #477, opened 2022-12-27
WIP: Fix other pages bugs

### LINKED ISSUE #471: Other pages bugs
- [x] About us page needs correct sponsors https://www-jsdelivr-com-pr-400.onrender.com/about (replace quantil with gcore)
- [x] ^ the same page has blurry logos of all sponsors
- [x] ^ the numbers must be real not static https://dl.dropboxusercontent.com/s/az1o71vzdo28ry2/chrome_2022-12-23_16-37-50.png
- [x] Network page needs an update in map locations. All CDNs have launched new POPs since then (and remove gcore ru locations since they are no longer accessible)
- [x] Stats page numbers need a different font like on homepage
- [x] Put Ghost 3rd https://www-jsdelivr-com-pr-400.onrender.com/custom-cdn-oss
- [x] In cache hit rates section, don't hardcode the providers list. Show the same providers as the API returned for the table above (after #460 e.g. quantil will be present in some 2021 period, so it should be in the cache hit rates section too). Take the values from `net-providers.json` https://dl.dropboxusercontent.com/s/f0ry0zobfqsmg32/2022-12-30_14-58-12.png
- [x] Dead scrollbar: https://dl.dropboxusercontent.com/s/pz9owwy5iqphbvd/2022-12-30_15-35-49.png - also probably restore the original size and show only three boxes with 4th hidden behind scroll

xbpcb:
closes #471


xbpcb:
Also, i will fix Stats page here, since the endpoints /v1/stats/network/countries and /v1/stats/network were modified

MartinKolarik:
Stats:
 - if a country filter is selected and the user clicks on a platform/browser table item, we end up in an invalid state - we need to select browser/platform breakdown in that case.
   - if "Group platform versions" / "Group browser versions" is disabled and the country filter is used, there is no breakdown option that could be used so clicking shouldn't do anything

xbpcb:
> Also, i will fix Stats page here, since the endpoints /v1/stats/network/countries and /v1/stats/network were modified

Fixed

MartinKolarik:
Merging, continue in a separate PR.

PR #491, opened 2023-01-22
Perf tests

jimaek:


src/views/components/header-gradient.html
```diff
@@ -2,7 +2,9 @@
 	<img class="header-gradient_primary"
 		width="1440"
 		height="684"
-		src="{{@shared.assetsHost}}/img/about/about-gradient.png" />
+		src="/img/about/about-gradient.png"
```
jimaek, line 5: because google wants to see it in next-gen format. So I could enable image optimization on the site level CDN and have this image auto-convert

PR #498, opened 2023-02-01
Add static date range options

### LINKED ISSUE #460: Add static date range options
Everywhere where a user is able to select a range like this:

image

Load https://data-jsdelivr-com-preview.onrender.com/v1/stats/periods in the background. Filter the results based on the stats type, e.g. when showing package stats, filter the results for `links.packages` (because not all data categories are available for all periods). Then add all available options to the dropdown, e.g.:

```
month
quarter
year
--- separator
Q3 2022
September 2022
August 2022
...
2021
December 2021
...
```

Endpoint docs: https://www-jsdelivr-com-pr-400.onrender.com/docs/data.jsdelivr.com#get-/v1/stats/periods

xbpcb:
Fixes #460


MartinKolarik:
Stats page:
- [ ] should be "yearly" https://dl.dropboxusercontent.com/s/cvp4e1i3aup9ezb/2023-02-01_16-16-38.png
- [ ] need an icon for stackpath (2021)
- [ ] if I switch to 2021 and there are no browser data available, the tables don't update - they should update to the "no data available for the selected period" message
  - [ ] same for the map and the chart above, when there are no data, show a message instead of keeping the old data

MartinKolarik:
Package page:
 - [ ] https://www-jsdelivr-com-pr-498.onrender.com/package/npm/mathjax?nav=stats chart not rendered for some reason but it renders always after a window resize https://dl.dropboxusercontent.com/s/7oy5c5sj0b5g9wz/2023-02-01_16-30-11.png
   - [ ] sometimes I also get a "no data" message even though the data is there, looks like a race condition somewhere in the requests https://dl.dropboxusercontent.com/s/4apykie03h96d7j/2023-02-01_16-33-54.png (but this one happens only randomly, not always)
 - [ ] "Stats for past January 2018" the word "past" must not be there if followed by a specific month/quarter/year.

MartinKolarik:
- [ ] Go to the homepage, search for react and open the package page. The page jumps down "Documentation" part of the readme for some reason.

src/public/js/_.js
```diff
@@ -758,14 +758,14 @@ module.exports = {
 
 		// create labels depending on chartPeriod, Screen size, groupBy
 		switch (chartPeriod) {
-			case 'week':
+			case 's-week':
```
MartinKolarik, line 761: `week` and `s-week` are two different types, so this change is technically wrong in most places. Instead, this should simply be:

```
case 'week':
case 's-week':
    results.labels = this.createWeekPeriodChartLabels(results.labels);
    break;

File: src/views/pages/oss-cdn.html [hunk 2]
```diff
@@ -55,7 +59,11 @@
 						{{/partial}}
 					</c-proxy-head>
 
-					<c-proxy-request-bandwidth-stats name="ghost" period="month" bandwidthUnit="GB"></c-proxy-request-bandwidth-stats>
+					<c-proxy-request-bandwidth-stats
+						name="ghost"
+						periodData="{periodValue: 'month', periodText: 'month', periodType: 's-month'}"
```
MartinKolarik, line 64: Do we actually need to pass around the period data like this? I would say this should stay as `period=month` everywhere, and only in the period dropdown can you add the text based on the period value.

src/views/components/package-sidemenu.html
```diff
@@ -6,18 +8,7 @@
...
-						<i class="fa fa-angle-down" aria-hidden="true"></i>
-					</button>
-
-					<ul class="dropdown-menu">
-						{{#each statsPeriods}}
-							<li><a on-click="@this.set('statsPeriod', @key)">{{@key}}</a></li>
-						{{/each}}
-					</ul>
-				</div>
+				<c-periods-dropdown selectedValue="{{periodData.periodValue}}" selectHandler="{{@this.handlePeriodSelecting()}}"></c-periods-dropdown>
```
MartinKolarik, line 11: Why do we need the select handler in addition to the selectedValue?

src/public/js/_.js
xbpcb: All these functions now expecting not value but type, since dropdown now has s-month type for all month-related values it is correct. I checked every case of usage, i think i didn't miss something.

src/views/pages/oss-cdn.html
xbpcb: Yes, because we use this component also on the _oss-cnd-project page

src/views/components/package-sidemenu.html
xbpcb: Handler we need to get back value to the parent, selectedValue is also used as a default value for the dropdown. Somewhere we have quarter as a default, somewhere month.

xbpcb:
>Go to the homepage, search for react and open the package page. The page jumps down "Documentation" part of the readme for some reason.
Hmm, can't reproduce it - checked in FF and Chrome - both act ok

MartinKolarik:
> Hmm, can't reproduce it - checked in FF and Chrome - both act ok

Happens only sometimes it seems, and also on master so moving to a separate issue: https://dl.dropboxusercontent.com/s/i8uh47d86jfifz5/2023-02-11_15-28-08.gif

PR #516, opened 2023-03-24
Add Globalping

xbpcb:
Add Globalping pages.
https://www-jsdelivr-com-pr-516.onrender.com/


xbpcb:
Not ready yet. Just a draft version.

MartinKolarik:
@xbpcb I updated the logo link logic, please update the CSS so that the "by jsDelivr" is again in the correct position.

jimaek:
2 new sponsors to add, ipinfo.io and ip2location.io. Maybe not next to each other :)

src/views/pages/globalping.html
```diff
@@ -1639,6 +1639,7 @@
 			}
 
 			function byteLinear (partOne, partTwo, quotient) {
+				// eslint-disable-next-line no-bitwise
 				let color = (('0x' + partOne) * (1 - quotient) + ('0x' + partTwo) * quotient) | 0;
```
MartinKolarik, line 1643: What is this meant to do? Bitwise or with `0` doesn't do anything.

xbpcb: It floors the number with handling NaN case, however yeah i think you are right and this time it is too much, replaced with regular Math.floor()

jimaek:
Issues:
Slack:
- [ ] Add hover state to GP menu
- [ ] Remove the empty space between selected item and header https://dl.dropboxusercontent.com/s/boywm1wzdu0oeqq/chrome_2023-04-19_18-45-00.png
- [ ] Background image gets sharply cut https://dl.dropboxusercontent.com/s/i0s72uh53qpvds6/chrome_2023-04-19_18-45-44.png

CLI:
- [ ] Background hex are stretched and blurry https://dl.dropboxusercontent.com/s/6a6ycvw95qs40zx/chrome_2023-04-19_18-47-49.png
- [ ] Both section header and text need update

Homepage:
- [ ] Add 2 sponsors
- [ ] A bit less whitespace would be better I think https://dl.dropboxusercontent.com/s/b75m3n0mn687sv4/chrome_2023-04-19_18-50-59.png
- [ ] Different font sizes https://dl.dropboxusercontent.com/s/pzhun7oyp99y04q/chrome_2023-04-19_18-52-01.png
- [ ] Fill the form is a link https://forms.gle/9jSzytJpTDYJmS2W8

Tests:
- [ ] By default show all probes in the hex style. Allow to click to zoom in
- [ ] After test runs clear the probes map and start showing the results as they come in. At the end zoom in to fit all the results, but not too much. e.g. if users asks to ping from Poland then we should zoom in to Poland, but not to fill 100% of the map. 
- [ ] The gear icon should only show a number equal to the amount of non-default settings https://dl.dropboxusercontent.com/s/zgosyzlyo43xtvi/chrome_2023-04-19_18-57-49.png If there are none then dont show anything
- [ ] The output should fit 3 ping results by default I think, right now its too small
- [ ] "See results" in tooltip must scroll down to the same result in the output below
- [ ] The side bar with range I think should be static. 5ms smallest green value and 200ms largest orange value. And the results can be of course outside the range, but the point is to show to the user the range of "good" latency and not the range of results
- [ ] Some tests can have multiple values, we should show all of them https://dl.dropboxusercontent.com/s/dwqhkt7t0xtrm3o/chrome_2023-04-19_19-03-20.png
- [ ] Enable real-time mode and show results as they come in with 1s GET loop

MartinKolarik:
When sharing the test results:
1. Add a message to the window (see screenshot)
2. Scroll down to the results on load
3. Change the button text from "Run Test" to "Run Again"
4. Use `?measurement`
image

MartinKolarik:
@xbpcb please continue with fixes in a new PR.

Issue #515, opened 2023-03-21
Add CDN alliance to history page


jimaek:
https://cdnalliance.org/

jimaek:
May 2023 - Add "Globalping goes live! Includes CLI and Slack App"
Link to /globalping

PR #527, opened 2023-05-02
Gh 524

### LINKED ISSUE #524: GP issues
Slack:
- [x] Add hover state to GP menu
- [x] Remove the empty space between selected item and header https://dl.dropboxusercontent.com/s/boywm1wzdu0oeqq/chrome_2023-04-19_18-45-00.png
- [x] Background image gets sharply cut https://dl.dropboxusercontent.com/s/i0s72uh53qpvds6/chrome_2023-04-19_18-45-44.png
- [x] When switching between Slack and CLI pages the background moves, it shouldnt

CLI:
- [x] Background hex are stretched and blurry https://dl.dropboxusercontent.com/s/6a6ycvw95qs40zx/chrome_2023-04-19_18-47-49.png
- [x] Both section header and text need update
- [x] Header should have a dynamic rounded number "global network of 500+ probes" in our case would be 600+
- [x] Some elements are broken on mobile

Homepage:
- [x] Add 2 sponsors
- [x] A bit less whitespace would be better I think https://dl.dropboxusercontent.com/s/b75m3n0mn687sv4/chrome_2023-04-19_18-50-59.png
- [x] Different font sizes https://dl.dropboxusercontent.com/s/pzhun7oyp99y04q/chrome_2023-04-19_18-52-01.png
- [x] Fill the form is a link https://forms.gle/9jSzytJpTDYJmS2W8
- [x] Make the forum link always underlined to make it clear its clickable
- [x] Sponsors is too close to contact us https://dl.dropboxusercontent.com/s/rse4vr8k7s86dpv/chrome_2023-04-21_20-57-01.png
- [x] Hide Network Tools page from menu
- [x] probe stats look broken: https://dl.dropboxusercontent.com/s/lz1w4g7tfst3rty/2023-05-09_14-32-51.png
- [x] footer broken
- [x] make the whole map/tests element wider, match the Join the network element size

Tests:
- [x] default protocol for HTTP should be HTTPS (docs were wrong about this)
- [x] Send the `limit` to the API as global limit, not per-location limit. (https://github.com/jsdelivr/globalping/blob/master/docs/measurement/schema/request.md#limit)
- [x] "limit" field must be large enough to show three digits, also "target" should be larger and "test type" as small as possible to fit the traderoute (limit still too small: https://dl.dropboxusercontent.com/s/zol8l2xtkj4g908/2023-05-09_14-23-17.png)
- [x] By default show all probes in the hex style. Allow to click to zoom in
- [ ] After test runs clear the probes map and start showing the results as they come in. At the end zoom in to fit all the results, but not too much. e.g. if users asks to ping from Poland then we should zoom in to Poland, but not to fill 100% of the map. 
- [x] The gear icon should only show a number equal to the amount of non-default settings https://dl.dropboxusercontent.com/s/zgosyzlyo43xtvi/chrome_2023-04-19_18-57-49.png If there are none then dont show anything
- [x] The output should fit 3 ping results by default I think, right now its too small
- [x] ^ but only if the results are >=3. In case they are fewer then size to fit https://dl.dropboxusercontent.com/s/a3u3zicxyuo8rbh/chrome_2023-04-24_19-04-44.png
- [x] "See results" in tooltip must scroll down to the same result in the output below
- [x] The side bar with range I think should be static. 5ms smallest green value and 200ms largest orange value. And the results can be of course outside the range, but the point is to show to the user the range of "good" latency and not the range of results
- [x] Some tests can have multiple values, we should show all of them https://dl.dropboxusercontent.com/s/dwqhkt7t0xtrm3o/chrome_2023-04-19_19-03-20.png
- [x] Enable real-time mode and show results as they come in with 1s GET loop
- [x] Dont group results in hex, thats only for probe listing. Results should always be individual dots.
- [x] Keep the same font between selected test and dropdownhttps://dl.dropboxusercontent.com/s/mz5g97ub2t2jzri/chrome_2023-04-25_19-28-11.png
- [x] Share results link should work and if opened in a new tab it should load everything in the same way and state
- [x] To make it easier to run a test I think we should pre-set some values. e.g target=cdn.jsdelivr.net location=world limit=3
- [x] When the user is in either input field then on enter run the test
- [x] Remove client-side parameter validation. Implement logic to correctly read all validation errors from the API and show these instead.
- [x] If there's an error in the options that are in the dropdown, we should open the dropdown, otherwise the error isn't visible when the user clicks "run":https://dl.dropboxusercontent.com/s/xm65hkemkqt4c2d/2023-05-09_14-37-36.png, also + margin between the text and the button below
- [x] There was supposed to be an "icon" with the number of set options I think?
https://github.com/jsdelivr/www.jsdelivr.com/assets/6192491/c5510e49-a950-4247-96de-e746ea4170f6

Other:
 - [x] Check all newly added SVGs. If they contain anything like this, they are NOT real SVGs, just PNGs encoded within SVG. Try to get real SVGs for the images. If not available, replace them with real PNG files. This "PNG in SVG" encoding doesn't help anything, just makes the file 30% bigger than plain PNG would be.
image
- [x] Bring back the GP menu item in top nav

xbpcb:
Fix #524


src/assets/less/components/header.less
```diff
@@ -148,7 +148,7 @@
 			margin: 12px 16px 12px 0;
 
 			@media (min-width: @screen-sm-min) {
-				margin: 15px 24px 15px 0;
+				margin: 15px 0;
```
MartinKolarik, line 151: This is wrong.

xbpcb: fixed

jimaek:
It's fine, with this branch merged we can have people visit the new pages

xbpcb:
Add commits for another task 
Fix #515

PR #539, opened 2023-05-18
Further Globalping improvement

### LINKED ISSUE #538: Further Globalping improvements
- [x] Dont parge the + symbol, the magic field accepts it and it acts as a filter
- [x] Instead use a comma to allow users to select multiple locations
- [ ]  After test runs clear the probes map and start showing the results as they come in. At the end zoom in to fit all the results, but not too much. e.g. if users asks to ping from Poland then we should zoom in to Poland, but not to fill 100% of the map.
- [ ] After the test is finished zoom in to fit all results. But not too much, something like a country level max (Same logic, zoom in to fit all dots on the map. Country is the max zoom in, but you must zoom out as much as needed)
- [ ] Only for HTTP tests: when the user pastes a value into target try parsing it with new URL() - if valid, set path, query, port, and protocol options based on those values, and open the options popup so that the users sees it. Example: I paste https://www.jsdelivr.com/package/npm/react?tab=files I get: https://dl.dropboxusercontent.com/s/i56l8rsvx27gbzo/2023-05-16_14-37-14.png
- [x]  for HTTP, use scale 5 ms -1000 ms (instead of 200 max) https://dl.dropboxusercontent.com/s/8xs0a5b7eeq573z/2023-05-16_21-00-27.png

xbpcb:
Fix #538

### DIFF
--- a/src/views/pages/globalping.html
+++ b/src/views/pages/globalping.html
@@ -1356,7 +1356,7 @@
 			reqParams.limit = limit || DEFAULT_LIMIT;
 
 			if (location) {
-				reqParams.locations = location.split('+').map(loc => ({ magic: loc }));
+				reqParams.locations = location.split(',').map(loc => ({ magic: loc }));
 			}
 
 			// since packets field default value is present only for Ping and MTR type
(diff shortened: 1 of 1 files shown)


jimaek:
I'll merge now since we're live and I want people to get fixes quickly for a better impression

Issue #542, opened 2023-05-22
Fix blog URLs 


jimaek:
- sitemap uses ghost.io https://www.jsdelivr.com/blog/sitemap-posts.xml

- Force canonical to always be www.jsdelivr.com

MartinKolarik:
@alexey-yarmosh the blog internally runs at https://jsdelivr-blog.ghost.io/ and we proxy it to www.jsdelivr.com/blog. The first problem is this (it should always be `www.jsdelivr.com`):
![image](https://github.com/jsdelivr/www.jsdelivr.com/assets/6192491/6aa8b85a-09bd-4cbd-9c6a-2146a78cd4f5)

Please check if we can do any of the following to:
 - customize how the link is output (change its value) in [our theme](https://github.com/jsdelivr/ghost-jsdelivr-theme),
 - remove the link in our theme,
 - add a "no index" meta tag in our theme - this one is certainly doable as the last option.

If we go with the third option, we can then simply remove that meta tag in our proxy.

The second issue, the sitemap:
 - again, if we can customize it via the theme, that would be easiest,
 - alternatively, our proxy can be adjusted to override values in XML too (the parser probably supports it), and rewrite the links in the sitemap the same way as we do it for HTML links: https://github.com/jsdelivr/www.jsdelivr.com/blob/master/src/proxy.js

alexey-yarmosh:
Canonical:
- it is not possible to configure the canonical value using admin panel or theme
- as it is a part of a `ghost_head` helper we can copy its output and edit/remove canonical there, but that is dirty
- there is no way to execute code on the server side, before it is sent to the user. And editing link using scripts on FE is too late
- adding a "no index" tag and remove it during proxying is the only option

Sitemap:
- no way to configure values there
- rewrite values on the proxy side seems to be the only option too

MartinKolarik:
@jimaek wdyt, adding no index to the ghost domain should mean canonical doesn't matter and solve the problem, right?

jimaek:
Yes if it works correctly it should be fine

MartinKolarik:
Ok so let's proceed with updating the proxy for both parts and then updating the theme to add the tag.

PR #551, opened 2023-05-31
fix: remove noindex tag & proxy sitemaps

alexey-yarmosh:
Fixes https://github.com/jsdelivr/www.jsdelivr.com/issues/542

Note. In this PR `harmon` breaks sitemap.xsl file, that need to be fixed in future:

image

### DIFF
--- a/src/proxy.js
+++ b/src/proxy.js
@@ -15,8 +15,9 @@ module.exports = (proxyTarget, host) => {
 	let hostUrl = new URL(host);
 	let proxyTargetPattern = new RegExp(`${_.escapeRegExp(proxyTarget)}/?`, 'gi');
 	let rewriteAttributes = [ 'action', 'content', 'href', 'link', 'src', 'srcset', 'style' ];
-	let rewriteElements = [ 'loc' ];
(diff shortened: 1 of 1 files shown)


MartinKolarik:
Causes some JS issues:
image

src/proxy.js
```diff
@@ -161,7 +170,7 @@ module.exports = (proxyTarget, host) => {
 						});
 				},
 			};
-		})), true),
+		})), false),
```
MartinKolarik, line 173: We certainly can't run this on everything, only HTML + XML.

alexey-yarmosh: true, I'll take it

MartinKolarik: 👍 we can probably keep it at `true` and do our own filtering based on response headers if there's no suitable option in the package directly

alexey-yarmosh: Here is the PR with a slightly another solution please check if it suits us. Seems to work as expected.
https://github.com/jsdelivr/www.jsdelivr.com/pull/553

PR #553, opened 2023-05-31
fix: noindex & proxy sitemaps v2

alexey-yarmosh:
Fixes https://github.com/jsdelivr/www.jsdelivr.com/issues/542

### DIFF
--- a/src/proxy.js
+++ b/src/proxy.js
@@ -169,7 +169,7 @@ module.exports = (proxyTarget, host) => {
 					});
 			},
 		};
-	})), false)
+	})), false);
 
 	return [
 		/**
--- a/src/proxy.js
+++ b/src/proxy.js
@@ -14,10 +14,6 @@ module.exports = (proxyTarget, host) => {
 	let proxyUrl = new URL(proxyTarget);
 	let hostUrl = new URL(host);
 	let proxyTargetPattern = new RegExp(`${_.escapeRegExp(proxyTarget)}/?`, 'gi');
-	let rewriteAttributes = [ 'action', 'content', 'href', 'link', 'src', 'srcset', 'style' ];
-	let rewriteElements = [ 'loc', 'image\\:loc' ];
-	let rewriteRegExp = [ 'script[type="application/ld+json"]' ];
-	let removeElements = [ 'meta[name="robots"][content="noindex"]' ];
 
 	let rewrite = (link, baseUrl) => {
 		// A relative URL without a leading slash. No transformation needed.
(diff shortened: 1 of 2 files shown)


src/proxy.js
```diff
@@ -102,66 +103,88 @@ module.exports = (proxyTarget, host) => {
...
-							} catch (e) {
-								stream.end(value);
-							}
-						});
-				},
-			};
-		})), true),
+		(req, res, next) => {
+			let path = req.path.toLowerCase();
+			let shouldParse = path.endsWith('/') || path.endsWith('.html') || path.endsWith('.xml');
```
MartinKolarik, line 180: Looks ok at first glance but would still prefer matching based on the content-type. You can use https://github.com/jshttp/type-is#typeisismediatype-types which is already internally used by express as well.

alexey-yarmosh: Here I am dealing with the `req`, but I can't find a way to get `req` headers including content-type, they are empty at that step.

alexey-yarmosh: I am able to get type inside `func` callbacks. But that will do a lot of useless parsing, plus `return;` should be replaced with something smarter to return the content without any changes. 
```js
rewriteAttributes.map((name) => {
		return {
			query: `[${name}]`,
			func (el, req, res) {
				const shouldParse = typeis.is(res._headers['content-type'], ['text/xml', 'text/html']);
				if (!shouldParse) {
					return;
				}

				let value = '';
				let stream = el.createStream()
					.on('error', () => stream.end())
					.on('data', chunk => value += chunk.toString())
					.on('end', () => {
						stream.end(value);
					});
			},
		};
	})
```

MartinKolarik:
Looks like we'll also need to rewrite:
 - xsl files (sitemap.xsl), which should be easy,
 - `robots.txt` but not other txt files, there a simple replace all via `rewriteAllAbsolute` should work - but this will need a check specifically for that one file. Alternatively, we could entirely mock the response for `robots.txt` and send content configured here.

jimaek:
Guys, spending 2 days on a sitemap proxy is not the best use of our time. Can we speed things up or compromise on something? I don't want to spend a 3rd day on it.

alexey-yarmosh:
Last things will be finished today or early tomorrow.

src/proxy.js
```diff
@@ -102,66 +169,40 @@ module.exports = (proxyTarget, host) => {
...
+		.concat(rewriteAttributesHTML.map(rewriteAttribute))
+		.concat(rewriteElementsHTML.map(rewriteElement))
+		.concat(rewriteRegExpsHTML.map(rewriteRegexp)), false);
+
+	let rewriteAttributesXML = [ 'href' ];
+	let rewriteElementsXML = [ 'loc', 'image\\:loc' ];
+	let harmonMiddlewareXML = harmon([], rewriteElementsXML.map(rewriteElement)
+		.concat(rewriteAttributesXML.map(rewriteAttribute)), false);
+
+	let rewriteAttributesXSL = [ 'href' ];
```
alexey-yarmosh, line 186: sitemap.xsl file breaks if we rewrite it's `content` attribute, thus we need to separate what to parse for what files.

MartinKolarik:
> Guys, spending 2 days on a sitemap proxy is not the best use of our time. Can we speed things up or compromise on something? I don't want to spend a 3rd day on it.

The proxy is a tricky business with this setup :)

PR #559, opened 2023-06-08
Gp: share results URL/CLI switcher 

### LINKED ISSUE #528: Share results/CLI switcher
Lets add a switcher here https://dl.dropboxusercontent.com/s/pza14homvxb0cxi/chrome_2023-05-02_13-48-04.png that changes the field from a URL to a CLI command that duplicates the test.

Default state would be same as now. It can contain "Share (icon) | CLI (icon)"

This way the user can for example continue using the same test in the terminal

xbpcb:
Fix #528


src/views/components/gp-test-results.html
```diff
@@ -155,5 +186,106 @@
...
+
+									if (params.measurementOptions.request.path) {
+										let path = params.measurementOptions.request.path;
+
+										path = path[0] === '/' ? path : `/${path}`;
+										str += ` --path ${path}`;
+									}
+
+									if (params.measurementOptions.request.query) {
+										str += ` --query "${params.measurementOptions.request.query}"`;
```
MartinKolarik, line 259: Use single quotes inside and add for `path` as well.

xbpcb: Done

PR #569, opened 2023-07-14
Fix Globalping issues

xbpcb:
Fix the issue with Markers and Clusters on the map (the reason was race condition)
Fix map scrolling into the view for Share-results-case
Some code improvements

### DIFF
--- a/src/views/pages/globalping.html
+++ b/src/views/pages/globalping.html
@@ -1170,8 +1170,8 @@
 				appliedMeasOpts: {},
 				shareResHeaderData: null,
 				runTestBtnText: 'Run Test',
-				shareResultsCase: false,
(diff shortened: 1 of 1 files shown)


src/views/pages/globalping.html
```diff
@@ -1874,8 +1855,14 @@
 
 					// should be after the mainOptions.type was set
 					this.set('realTimeTestResResponse', testRes);
-				}).catch(() => {
+				}).catch((e) => {
 					clearInterval(testReqInterval);
+
+					if (e && typeof e === 'string' && e.toLowerCase() === 'not found') {
```
MartinKolarik, line 1861: This will likely break soon as we change the message format. The check should be for 404 status, not the text.

PR #573, opened 2023-08-03
Set probe as a location

### LINKED ISSUE #536: GP: Set probe as location
https://dl.dropboxusercontent.com/s/tu5492d5dvany40/chrome_2023-05-15_18-20-02.png

Lets add a second line under show results called "Set as location"
It will then overwrite the current location input and set the probe using something like "asn+city"

e.g. in the screenshot the location would be "AS40676+Johannesburg"

This way the user can continue the journey more easily. e.g. I would run a traceroute from that location to understand why the latency is show high

xbpcb:
Fix #536


MartinKolarik:
How about https://fontawesome.com/v4/icon/hand-o-right ?

jimaek:
Doesn't look good to me. Something tagged as add or location should work https://fontawesome.com/v4/icons/

MartinKolarik:
https://fontawesome.com/v4/icon/map-marker then?

jimaek:
Works for me

MartinKolarik (commented):
@xbpcb please update the icon and also the position on mobile: https://dl.dropboxusercontent.com/s/cbq8cf0dqf4j086/2023-08-06_17-16-46.png

It should stay on the left, same as on larger layouts.

PR #579, opened 2023-08-24
Add limit/pagination to top package versions/files and files stats to oss cdn

### LINKED ISSUE #507: Add custom limit and pagination to top package versions/files
- add to: https://dl.dropboxusercontent.com/s/dvz8nlbhim24n2z/2023-02-18_00-02-40.png
 - same design as "Top platforms" on Stats page
 - the selected limit for versions in "Top 5 versions" table should also apply to "Requests per quarter per version" chart below
 - change the titles to "Top versions" and "Top files" (remove the numbers)
 - related task: #508

### LINKED ISSUE #508: Add files stats to oss cdn pages
- below the existing chart, add this table https://dl.dropboxusercontent.com/s/7cvknjitldxrjyj/2023-02-18_00-06-19.png but full-width (there are no versions in oss cdn so we only have one table to show and the filenames are going to be long)
- API endpoint: https://www.jsdelivr.com/docs/data.jsdelivr.com#get-/v1/stats/proxies/-name-/files
- #507 applies here as well

xbpcb:
Fixes #507 
Fixes #508


MartinKolarik:
- [ ] top versions/files - when I change the number of records per page, I want to see all of them, there should be no scroll bar; to make it look good, when the number is changed in one table, change it in the other as well so that they both stay the same height
 - [ ] oss cdn - also should have no scroll bar inside the table
 - [ ] Requests per quarter per version - looks ok, just see if we can come up with some more different colors - right now some are very similar

MartinKolarik (approved):

PR #587, opened 2023-10-03
Use jQuery slim version

xbpcb:
Implement jquery.slim.js and fix everything related (Main page: search, scroll to results; package page: search-compact, etc.)

### DIFF
--- a/src/views/pages/_package.html
+++ b/src/views/pages/_package.html
@@ -240,12 +240,5 @@
 				}
 			}
 		},
-		onrender () {
-			if (!Ractive.isServer) {
-				// make sure scrollY is at 0 whne page is rendered
-				window.scrollTo(0, 0);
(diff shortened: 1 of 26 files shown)


src/views/pages/_package.html
```diff
@@ -240,5 +240,12 @@
 				}
 			}
 		},
+		onrender () {
+			if (!Ractive.isServer) {
```
MartinKolarik, line 244: This seems wrong. Why do we need it? If the browser remembers a scroll position after e.g. page refresh, it shouldn't jump back to the top.

xbpcb: You're right, this was supposed to prevent the page from incorrect position after navigating from the search results, but I missed that it would work every time, even when the page reloaded. Fixed.

PR #590, opened 2023-10-07
Link to doc

jimaek:


jimaek:
@xbpcb Please check the color of the links. We need something that feels more native with the design

PR #607, opened 2023-11-13
Terms

MartinKolarik:
Currently loads the documents from the `terms` branch, the links will need to be updated after the docs are merged. Also, links within the text won't work in the preview as they link to the production domain.


jimaek:
Lets also replace the contact email dak@prospectone.io with d@jsdelivr.com

PR #609, opened 2023-11-21
GP: Add more details to the map after zooming

### LINKED ISSUE #608: GP: Add more details to the map after zooming
Need to solve the problem of "empty grey map" after zooming and making it impossible to understand where you are. Lets add more details to the map after a few levels of zoom.
Lets test different versions.

xbpcb:
Fix #608

### DIFF
--- a/src/views/pages/globalping.html
+++ b/src/views/pages/globalping.html
@@ -1005,7 +1005,7 @@
 </r-page>
 
 <script>
-	const styles = require('../../assets/js/map/styles.json');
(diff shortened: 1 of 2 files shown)


MartinKolarik:
I wonder, can we force Google Maps to use English regardless of the browser setting? It's weird when everything is in English, but the map uses my local language.

jimaek:
If its simple, sure lets add it. This level of details looks good btw

PR #626, opened 2024-03-30
Fixed Gcore logo height

coozywana:
Before
318193016-10f8b65e-29b9-46b1-a899-7d25b3731c06
After
Screenshot 2024-03-30 at 11 33 07 pm

Made the height 100px as all the other logos in the sponsors section are.

https://github.com/jsdelivr/www.jsdelivr.com/assets/73876112/a0993cd9-2189-4934-8404-e65345edc162

Gcore is the only one with 116px


MartinKolarik:
Hi, thanks for the PR, but I don't think this change makes sense here. The logo is an SVG, and the change even makes the file bigger.

PR #645, opened 2024-06-26
GP: infinite mode totals and seq

### LINKED ISSUE #637: GP: infinite mode totals and seq
*Separate PR after merge of #624*

Infinite mode output in raw mode is missing this part (should appear after it stops):
```
--- cdn.jsdelivr.net.cdn.cloudflare.net ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 1002ms
rtt min/avg/max/mdev = 5.614/5.630/5.652/0.016 ms
```
See these related comments for implementation details: 
 - [ ] https://github.com/jsdelivr/globalping-cli/pull/86#discussion_r1468153736
 - [ ] https://github.com/jsdelivr/globalping-cli/pull/86#discussion_r1486035142
 - [ ] https://github.com/jsdelivr/globalping-cli/pull/86#discussion_r1476013293

Also, `seq` values in the raw output must be rewritten on the client side to be continuous as discussed in Slack.

xbpcb:
Fix #637

### DIFF
--- a/src/views/pages/globalping.html
+++ b/src/views/pages/globalping.html
@@ -3062,6 +3062,7 @@
 							areTimingsReady: this.checkAreTimingsReady(lastTiming, maxTiming, minTiming, avgTiming, packetsTotalCombined),
 							rawOutput: this.handleRawOutput([ linesHeader, ...outputLinesCombined ]),
 							extraValues: lossValue !== null ? { loss: { text: 'Loss', value: lossValue, units: '%' } } : {},
+							rawFooter: this.createRawFooter(probeData.target, packetsTotalCombined, numericPacketsRttCombined.length, lossValue, minTiming, avgTiming, maxTiming),
 						}],
 					};
 				}, {});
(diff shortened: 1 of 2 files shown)


src/views/pages/globalping.html
```diff
@@ -3135,7 +3174,9 @@
 
 				if (lines[i].includes('---')) { break; }
 
-				packetLines.push(lines[i]);
+				let numberedLine = lines[i].replace(/icmp_seq=\d+/, `icmp_seq=${idxShift + i}`);
```
MartinKolarik, line 3177: I think this won't work correctly for this case:
```
no answer yet for icmp_seq=1
64 bytes from api.globalping.io (49.12.17.170): icmp_seq=1 ttl=51 time=15.9 ms
no answer yet for icmp_seq=2
no answer yet for icmp_seq=3
64 bytes from api.globalping.io (49.12.17.170): icmp_seq=3 ttl=51 time=15.3 ms
no answer yet for icmp_seq=4
```

Sum the shift with the actual number from the regex match because the numbers can repeat.

PR #655, opened 2024-07-17
GP: Improve HTTP target handling

### LINKED ISSUE #648: GP: Improve HTTP target handling
The same logic that runs after pasting to the target field should also run when the target is "finished" (blur event or tab/enter to start the second target).

Paste `https://cdn.jsdelivr.net/npm?x=1` - good:

image

Input `https://cdn.jsdelivr.net/npm?x=1` and hit tab - bad:

image

xbpcb:
Fix #648

### DIFF
--- a/src/assets/less/components/tags-input.less
+++ b/src/assets/less/components/tags-input.less
@@ -67,7 +67,6 @@
 		box-shadow: 0 2px 2px rgba(0, 0, 0, .04);
 
 		@media (min-width: @screen-sm-min) {
-			overflow-x: unset;
 			height: 40px;
 			flex-direction: row;
 			justify-content: flex-end;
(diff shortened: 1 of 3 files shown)


src/assets/less/components/tags-input.less
```diff
@@ -67,7 +67,6 @@
 		box-shadow: 0 2px 2px rgba(0, 0, 0, .04);
 
 		@media (min-width: @screen-sm-min) {
-			overflow-x: unset;
```
MartinKolarik, line 70: This was here for a reason.

image

xbpcb: Hm, true, but it breaks styles in other way. I'll check it again.

xbpcb: Should be OK now

PR #660, opened 2024-08-06
GP: IPv6 support

### LINKED ISSUE #643: GP: IPv6 support in the measurement tool
- [x] all fields that support IP addresses must accept IPv6 format (target + resolver in options)
 - [x] there's a [new switch](https://dl.dropboxusercontent.com/scl/fi/4qk30ra4yzu5u9dlkbzdt/2024-06-24_19-45-52.png?rlkey=hwh018a1c7yfzjf2d3kuwpz1f&dl=0) in the target field (see Figma), default is IPv4
     - [x] when the user enters IP, automatically switch based on the IP format
 - [x] the switch value is sent as `measurementOptions.ipVersion: 4 | 6` to the API, but *only if* the target is a domain
 - [x] the switch should be hidden for DNS test type

xbpcb:
Fix #643


MartinKolarik (commented):
Looks good except one small thing. While the measurement runs, all input fields are being disabled, but this one isn't.
image

MartinKolarik:
Unrelated to the PR but let's fix: the white dot in the switch moves on focus: https://dl.dropboxusercontent.com/scl/fi/cnje6giojv54j2icp6xs9/2024-08-06_21-37-27.gif?rlkey=xj67a7oatapp89lp32sxdtdzz&dl=0

MartinKolarik:
Actually a few more:
 - [ ] infinite mode in table view doesn't show IPv6 addresses,
 - [ ] regular mode in table view overflows with IPv6 addresses - not sure what to do here as the addresses can be quite long; maybe we don't need to show them at all @jimaek? (cli doesn't in table mode)
 - [ ] if I select IPv6, then change the target (to another domain), it shouldn't reset to IPv4.

MartinKolarik:
@xbpcb I'll deal with the overflows, just do the remaining two points.

MartinKolarik (approved):
Going to merge after jsdelivr/globalping-probe#241 is fixed.

PR #663, opened 2024-08-08
GP: switching test type with map off

### LINKED ISSUE #662: GP: switching test type with map off
1. Run a measurement, disable map:
 
image

2. Switch type. The results disappear, and it looks weird:

image

It would be good to show the map again after step 2, then hide it when we start showing results for the second measurement (right when this part appears):

image

xbpcb:
Fix #662


MartinKolarik:
This doesn't look very good in the end :( Is there a reason why we hide the results after changing the test type? Editing other fields doesn't do anything with the results until the user hits Run. Maybe we can just keep the results there and not mess with the map.

PR #664, opened 2024-08-14
GP: clicking around inputs, IP switch styles

### LINKED ISSUE #661: GP: clicking around the target field
The target field is inside a `label` with padding, which means clicking anywhere on the whitespace inside focuses the field, which is very confusing. Clicking outside the input shouldn't do anything. Not sure about the reasons, maybe we can just replace the wrapper with a div and keep the label separately (not as a parent)?

image

Update: this applies to location and limit fields too.

xbpcb:
Fix #661


MartinKolarik:
> Also - when focusing the target field, the IPv selector moves. It shouldn't.

Fixed on desktop but still happens on mobile.

MartinKolarik (approved):

PR #674, opened 2024-08-16
GpRawOutput: fix status bars colors logic

### LINKED ISSUE #668: GP: measurement results doesnt show node statuses correctly
Run any measurement type except the "ping" one and you'll see that when the test is not failed or offline there will be no success status bar (green one) but only an in-between status (grey one).
Same goes to the Network-tools page too

xbpcb:
Fix #668

### DIFF
--- a/src/assets/js/_.js
+++ b/src/assets/js/_.js
@@ -1503,4 +1503,48 @@ module.exports = {
 			return cache.get(key);
 		};
 	},
+	getGpProbeStatusColor (timing, probesMaxTiming = 200, probesMinTiming = 5) {
+		// return default GREY color while probe has no timing yet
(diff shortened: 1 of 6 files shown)


MartinKolarik:
I think the way this "status line" works is confusing (before this PR as well) as the UI doesn't say what the colors mean and we use the same colors as in the map above for different things. I suggest we change it to use the same logic as the markers on the map:
 - in progress - light gray
 - success - color scale based on the value
 - error, offline, whatever - dark

This way the user will more likely match the meaning with the map.

@jimaek what do you think? 

image

image

jimaek:
Makes sense :)

MartinKolarik (approved):

PR #683, opened 2024-09-19
GP: Credits page

### LINKED ISSUE #673: GP: credits
Design in Figma.

xbpcb:
Fix #673


jimaek:
- Need to fix the slider
- Lets update the values to proper ones. @MartinKolarik I believe the excel doc is accurate, right?
- "Host a probe" probably lets link to /#:~:text=Our%20users%3A-,Join%20the%20network,-To%20join%20the
- The fonts seem all wrong, different sizes everywhere, too small in other places. I think its best to use the one within the explanation sections

MartinKolarik (commented):
Slider issues:
 - [ ] the handle can be moved entirely outside the range
 - [ ] the value should update as I'm dragging it (not only after I stop)
 - [ ] the $$ values below are not aligned very well

Misc:
 - [ ] link colors in FAQ
 - [ ]

MartinKolarik:
Looks like the slider also:
 - breaks itself on window resizing (wrong handle position),
 - overflows on windows resizing when previously dragged to the right,
 - doesn't respond to touch events at all.

Maybe it would be a good idea to look into some existing lib/component for this.

MartinKolarik:
Or maybe simply use the range input with a bit of extra markup around it: https://www.smashingmagazine.com/2021/12/create-custom-range-input-consistent-browsers/

PR #709, opened 2024-11-15
Add GP terms

MartinKolarik:


jimaek:
FYI when loading fails it points to the wrong repo https://share.perfstack.net/2024/11/chrome_2024-11-16_11-01-42.png
But I guess that should never happen in reality

PR #711, opened 2024-12-02
Migrate to new Google Map Markers, improve GP main page clustering

### LINKED ISSUE #635: GP: probes clustering on the map
I noticed the coordinates we're getting from geoip providers are very low quality and tend to put all probes in the area in a single place. E.g. [this](https://dl.dropboxusercontent.com/scl/fi/fqds5qa1u8zti96p8cgrp/2024-06-17_23-30-46.png?rlkey=7ygcz1am0tniljx8ijznohi7j&dl=0) changes to [this](https://dl.dropboxusercontent.com/scl/fi/ue2gs9nsy4198jcta7ga0/2024-06-17_23-32-41.png?rlkey=csl7mvus54s3qguqql3etvhbe&dl=0) when zooming, which is 1 probe on the left, and 17 probes stacked over each other on the right. In some other cases, it results in formations like [this](https://dl.dropboxusercontent.com/scl/fi/3ul1loxbckhtw9cyo3d8g/2024-06-17_23-33-51.png?rlkey=k9jl7uai53efxau2xwelyfc94&dl=0). 

We have a somewhat related task https://github.com/jsdelivr/globalping/issues/306 (this question is about UI, but it might create new requirements in that task), and I'm wondering what's even the best way of presenting these on the map:
 a) slightly moving the dots so that each individual probe is visible (similar to the last screenshot),
 b) forcing the cluster view and always showing a single node with the number,
 c) something else?

I think B makes the most sense (with 20 dots, you aren't going to click each of them to see the details anyway), and it allows for the most efficient API responses, but it opens the question of what to do about tooltips with network names - do we just not show them for these clusters? Do we show a big tooltip with all network names?

@jimaek @alexey-yarmosh

### LINKED ISSUE #638: Migrate to the new Google Map Markers

xbpcb:
Fix #638 
Fix #635


src/views/pages/globalping/_index.html
```diff
@@ -2158,65 +2121,65 @@
...
 				renderer: {
-					render: ({ markers, _position: position }) => {
-						return new google.maps.Marker({
-							position: {
-								lat: position.lat(),
-								lng: position.lng(),
-							},
-							icon: `${this.get('@shared.assetsHost')}/img/globalping/map_cluster.svg`,
-							label: { text: String(markers.length), color: '#fff', fontSize: '16px', fontWeight: '600' },
+					render ({ markers, position }) {
```
MartinKolarik, line 2148: The text doesn't seem to be vertically centered: https://dl.dropboxusercontent.com/scl/fi/jqnhkie2htwrnfenx26xb/2024-12-03_15-52-34.png?rlkey=xevkh5v8aoo3ptnpucucaiuol&dl=0

xbpcb: fixed

PR #716, opened 2024-12-17
GP: Duplicate meta descriptions

### LINKED ISSUE #715: GP: Duplicate meta descriptions
https://globalping.io/network-tools/http
https://globalping.io/terms

It seems that all of our pages have duplicate meta descriptions, we need to fix that and have them all be unique. Let me know what content is needed and I will provide it

xbpcb:
Fix #715


xbpcb:
In-progress.
There is an issue with Network-tools meta descriptions and Google.
I'm working on it.

src/views/pages/globalping/network-tools.html
```diff
@@ -223,6 +223,36 @@ <h1>{{displayingTestType}} from multiple locations in </h1>
 	const PROBE_STATUS_OFFLINE = _.getProbeStatusOfflineValue();
```
MartinKolarik, line 223: What exactly was broken on these pages and how does it change?

src/views/pages/globalping/network-tools.html
```diff
@@ -223,6 +223,36 @@ <h1>{{displayingTestType}} from multiple locations in </h1>
...
 
 	component.exports = {
+		computed: {
+			titleAndDescr () {
```
MartinKolarik, line 227: I don't understand why this was moved to be a computed property if it's still used just once during the load.

xbpcb:
There's another problem - we have several pairs of different URLs pointing to the same page, such as /ping and /ping-from-world, /http and /http-from-world, and so on.
They share the same meta description which might cause the same issue with Google.

MartinKolarik (changes requested):
`/network-tools/http-from-amazoncom-inc` doesn't work now.

src/routes/globalping.js
```diff
@@ -55,7 +55,7 @@ koaElasticUtils.addRoutes(router, [
  * Network tools pages.
  */
 koaElasticUtils.addRoutes(router, [
-	[ '/network-tools', '/network-tools/:params?' ],
```
MartinKolarik: This was ok before, the first value is just a name reported in APM.

MartinKolarik (approved):

PR #718, opened 2024-12-17
fix: extra space in package url

lubieowoce:
Potential fix for #18608. Not sure why the extra `<span>` is useful in the first place.


lubieowoce:
Already fixed in https://github.com/jsdelivr/www.jsdelivr.com/commit/aa63589e39af09637d91224dd01cf3c4230d4804

MartinKolarik:
Looks good. Indeed I don't see the need for that span now. Possibly it's a leftover from something older.

lubieowoce:
uh oh, i guess that `span` was needed to update things client-side? because that's broken now.

https://github.com/user-attachments/assets/61f37ba9-e1d4-4d19-9195-99b7902c94eb

PR #721, opened 2025-01-07
Gp: fix tooltips positions on mobile screens

xbpcb:
Also, apply unified styles for all GP tooltips


MartinKolarik (commented):
The positioning is better, but the padding should stay as it was on short tooltips (even if there's a change they'll break in two lines). Only those that originally had the multiple attribute should have the bigger padding.

PR #732, opened 2025-03-05
Load Fonts From jsDelivr

SukkaW:
Load fonts `Inter` and `Lexend` on `cdn.jsdelivr.net` from packages published by `@fontsource`.

Note that the `Source Code Pro` and `Open Sans` fonts are still loaded on Google Fonts since I can't find `Source Code Pro` with a weight of 500 and `Open Sans` with a weight of 200 on npm.

### DIFF
--- a/src/views/r-docs-globalping.html
+++ b/src/views/r-docs-globalping.html
@@ -61,7 +61,10 @@
 <!--			<link rel="preconnect" href="https://datum.jsdelivr.com/" crossorigin="anonymous">-->
 			<link rel="preconnect" href="https://fonts.googleapis.com" crossorigin="anonymous">
 
-			<link href="https://fonts.googleapis.com/css2?family=Lexend:wght@400;600;700&display=swap" rel="stylesheet">
+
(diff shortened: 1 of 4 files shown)


SukkaW:
cc @MartinKolarik

MartinKolarik:
Thanks, but I don't think this improves anything as it results in many more requests compared to Google Fonts, and if we still keep Google Fonts for other font families, even stuff like "more privacy" or "fewer DNS queries" doesn't apply.

SukkaW:
> Thanks, but I don't think this improves anything as it results in many more requests compared to Google Fonts, and if we still keep Google Fonts for other font families, even stuff like "more privacy" or "fewer DNS queries" doesn't apply.

The idea of the PR is not about fewer DNS queries, few connections, or privacy stuff. The idea is as simple as "Let's use jsDelivr more and see how we can migrate the rest".

SukkaW:
Also, it does make sense for jsDelivr-owned websites to use more jsDelivr anyway.

PR #733, opened 2025-03-10
Add autocomplete for location field

### LINKED ISSUE #537: GP: autocomplete for location field
Details TBD; need to consider tags handling as well.

MartinKolarik:
Referencing US states and cities requires https://github.com/jsdelivr/globalping/issues/629, and doesn't work at the moment. Closes #537.

Also fixes displaying error messages for the location field and hides the map in case of errors that are shown below it, as those were easy to miss otherwise. 

TODO:
 - [x] If two values are present, autocomplete should not delete the first one.
 - [x] Improve `+` to work with networks and tags.
 - [x] Try to add a "no results" message.

### DIFF
--- a/elastic-apm-node.js
+++ b/elastic-apm-node.js
@@ -1,7 +1,7 @@
 module.exports = {
 	active: process.env.NODE_ENV === 'production',
 	serviceName: process.env.SITE === 'globalping' ? 'globalping-website' : 'jsdelivr-website',
-	serviceVersion: require('./package.json').version,
+	serviceVersion: process.env.RENDER_GIT_COMMIT || require('./package.json').version,
 	logLevel: 'fatal',
 	centralConfig: false,
 	captureExceptions: false,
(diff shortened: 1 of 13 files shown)


MartinKolarik:
Not sure how to make it clearer than this, it's the exact reason that message is there:
image

Also:
 - network names are supported, only AS numbers are not,
 - system tags are supported (user tags not as they'd add too much noise, I think),
 - `+` is supported (it's even used by the autocomplete itself if you select cities for example).

MartinKolarik:
> + is supported (it's even used by the autocomplete itself if you select cities for example)

I guess you meant a combination of network/tag and a place? That it indeed doesn't work but I think I can make that work too.

jimaek:
Yes a combo, that's what I use the most of all combos.

How about showing a message when no matches are found? That's when it's needed most. Maybe show the contents of the tooltip to make it clear that lack of results doesn't mean it's not going to work

MartinKolarik:
> How about showing a message when no matches are found? That's when it's needed most. Maybe show the contents of the tooltip to make it clear that lack of results doesn't mean it's not going to work

The tooltip message won't directly help, but I was thinking in that direction, too, now. Will try to write something to show in that case.

MartinKolarik:
@jimaek check the `+` handling behavior now.

PR #740, opened 2025-03-24
Load Fonts From jsDelivr

SukkaW:
Follows #732. Now all Google Fonts' references are replaced with jsDelivr.

### DIFF
--- a/src/views/pages/osscdn.html
+++ b/src/views/pages/osscdn.html
@@ -5,7 +5,10 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1">
 		<meta name="description" content="jsDelivr - A free, fast, and reliable Open Source CDN for npm and GitHub">
 		<title>OSSCDN.com is now jsDelivr</title>
-		<link rel="stylesheet" href="//fonts.googleapis.com/css?family=Open+Sans:200,400,600,700">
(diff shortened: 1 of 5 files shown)


src/views/pages/osscdn.html
```diff
@@ -5,7 +5,10 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1">
 		<meta name="description" content="jsDelivr - A free, fast, and reliable Open Source CDN for npm and GitHub">
 		<title>OSSCDN.com is now jsDelivr</title>
-		<link rel="stylesheet" href="//fonts.googleapis.com/css?family=Open+Sans:200,400,600,700">
+		<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@fontsource/open-sans@5.2.5/400.css">
```
SukkaW, line 8: `Open Sans` doesn't have a weight of 200. The Google Fonts does not have as well.

MartinKolarik:
I wonder if it wouldn't be better to use the variable fonts via our special endpoint instead (see https://fontsource.org/fonts/lexend/cdn). The CSS could be inlined, and the single font file is smaller than 3 separate ones. Is there a specific reason you chose the static files?

SukkaW:
> Is there a specific reason you chose the static files?

Hmmmm, not exactly. The npm endpoint was chosen because it was the first to pop up. Lemme check if I can migrate to variable fonts!

PR #751, opened 2025-04-04
Sort version list by stability

### LINKED ISSUE #601: Improve version dropdowns
The version dropdowns, as they are now, are quite messy for projects that publish prereleases, e.g.:
image

How about we split the version list into three sections with a separator in between:
1. All stable, non-deprecated versions.
2. All prereleases.
3. Any deprecated versions.

Each section would be sorted on its own from the most recent.

mpourismaiel:
Closes #601 

Keeps the order from api and only applies the feature to non-github packages, let me know if it should apply there as well (assuming no since those are sha versions).

Additional features added:
- Keep separator when searching
- Add extra padding to no versions found message

### DIFF
--- a/src/views/components/version-dropdown.html
+++ b/src/views/components/version-dropdown.html
@@ -53,19 +53,18 @@
 						}, [ ...versions ]);
 				}
 
-				for (let i = 0; i < results.length - 1; i++) {
-					if (results[i] === 'separator' && results[i + 1] === 'separator') {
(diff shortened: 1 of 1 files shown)


MartinKolarik:
@mpourismaiel now that you have write access to the repos, please send future PRs from branches directly in this repo, not from your fork. That way we get automated preview deployments.

MartinKolarik (approved):

PR #754, opened 2025-04-06
Make package file section more accessible

### LINKED ISSUE #685: Clickable elements of package file browser & copy dropdown aren't accessible
Screenshot of the package file browser taken with tabbing order annotations enabled in the "Accessibility" tab of Firefox devtools:

<br>

**We should probably add:**

1. The [breakcrumb](https://github.com/jsdelivr/www.jsdelivr.com/blob/af62667a8b8449118b722aaa9096bc6a2ef47a67/src/views/components/package-file-browser.html#L8), [parent directory](https://github.com/jsdelivr/www.jsdelivr.com/blob/af62667a8b8449118b722aaa9096bc6a2ef47a67/src/views/components/package-file-browser.html#L76), [directory](https://github.com/jsdelivr/www.jsdelivr.com/blob/af62667a8b8449118b722aaa9096bc6a2ef47a67/src/views/components/package-file-browser.html#L88) and [show more](https://github.com/jsdelivr/www.jsdelivr.com/blob/af62667a8b8449118b722aaa9096bc6a2ef47a67/src/views/components/package-file-browser.html#L150) links aren't keyboard accessible.
   - These are already `<a>`s, but without `href`s. Specify `tabindex="0"` explicitly.
2. The copy to clipboard buttons (and therefore the [copy dropdown](https://github.com/jsdelivr/www.jsdelivr.com/blob/af62667a8b8449118b722aaa9096bc6a2ef47a67/src/views/components/copy-dropdown.html)) aren't keyboard accessbile.
   - [Button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) role for the trigger.
   - [Listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) roles and [popup](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup) attributes for the menu & and its items.
   - Key press handlers for the menu items (<kbd>↑</kbd> & <kbd>↓</kbd> to switch items, <kbd>Esc</kbd> to close the menu).
3. The [add to collection switch](https://github.com/jsdelivr/www.jsdelivr.com/blob/af62667a8b8449118b722aaa9096bc6a2ef47a67/src/views/components/package-file-browser.html#L127) isn't keyboard accessible.
   - [Toggle button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#toggle_buttons) role for the `<label>`.

mpourismaiel:
Closes #685

I changed some anchor tags which didn't have href to button. In my testing there wasn't any issues and tabindex was working properly but I'm not sure if there was an explicit reason for the anchor instead of button. We can add event listeners to anchors to simulate button if necessary.

### DIFF
--- a/src/views/components/copy-dropdown.html
+++ b/src/views/components/copy-dropdown.html
@@ -60,10 +60,6 @@
 						items.eq(nextIndex).focus();
 					}
 				});
-
-				setTimeout(() => {
-					dropdown.find('.dropdown-menu li:first-child button').focus();
-				}, 10);
 			};
 
 			let handleDropdownHidden = (event) => {
--- a/src/views/components/copy-dropdown.html
+++ b/src/views/components/copy-dropdown.html
@@ -1,25 +1,25 @@
 <div class="c-copy-dropdown btn-group {{#if dropup}}dropup{{/if}}">
-	<span data-toggle="dropdown">
+	<button data-toggle="dropdown" role="button" aria-haspopup="true">
 		<img width="20" height="20" src="{{@shared.assetsHost}}/img/icons/copy-btn-black.svg" loading="lazy"/>
-	</span>
+	</button>
 
-	<ul class="dropdown-menu dropdown-menu-right">
+	<ul class="dropdown-menu dropdown-menu-right" role="listbox" aria-label="Copy options" aria-expanded="false">
 		<li>
-			<a as-clipboard="undefined, dropup ? 'bottom' : 'top', '.c-copy-dropdown'" data-clipboard-text="{{url}}">Copy URL</a>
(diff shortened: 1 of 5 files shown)


src/views/components/copy-dropdown.html
```diff
@@ -41,5 +41,44 @@
...
+							nextIndex = items.length - 1;
+						}
+
+						items.eq(nextIndex).focus();
+					}
+				});
+
+				setTimeout(() => {
+					dropdown.find('.dropdown-menu li:first-child button').focus();
+				}, 10);
```
MartinKolarik, line 66: This triggers focus and, hence, a visual outline if I open the dropdown with the mouse, which is wrong. But it seems everything works fine without it as well, just requires one more tab?

PR #789, opened 2025-07-08
Fix quick filters in the network page

mpourismaiel:
I hadn't considered quick filters when fixing an issue with location input. The only fix I could think of to avoid a regression was to add another 1-way bound variable.

With this the setup is as follows:

- `overrideLocation` passes value to location-input
- `location` receives value from location-input

And these are used as:

- `location` for updating url and `filtersSet`
- `overrideLocation` for passing the quick filters changes and initial url variables to location-input

### DIFF
--- a/src/views/pages/globalping/network.html
+++ b/src/views/pages/globalping/network.html
@@ -242,7 +242,7 @@
 				_,
 				title: 'Network - Globalping',
 				description: 'Explore the global network map of Globalping probes.',
-				location: '',
+				location: 'World',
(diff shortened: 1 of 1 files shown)


MartinKolarik:
> Also I think we should not move the order of quick filters. When clicking on a filter it moves to the front and its confusing as the whole list changes. Just enable them in place.

It is a bit annoying, but it's how it worked before, and they'd move even if we kept them in place because of the extra "X" that appears when a filter is active.

PR #855, opened 2026-05-19
fix: do not catch errors when fetchStats fails in og image mw

PavelKopecky:
As discussed on Slack


PavelKopecky:
@coderabbitai full review

